PR #1, opened 2017-08-21
Initial version of the gem with Sentry support

tijmenb:
This gem is meant to be used by all GOV.UK Ruby applications. 

Once included it will configure the application to use Sentry for error reporting.

## What does it do?

- Pull in `sentry-raven`, the Sentry client for Ruby
- Configures Sentry's `current_environment`
- Before all error reports, increment a statsd counter (this is what we currently have in Errbit)
- When an error report fails, increment another statsd counter so we can be notified of Sentry failures
- Exposes a `GOVUK::Error` abstraction to manually report errors
- Exposes the statsd client so the app can use it

## Testing

To show how this works in real life these PRs have been raised:

- https://github.com/alphagov/content-tagger/pull/422
- https://github.com/alphagov/whitehall/pull/3416
- https://github.com/alphagov/collections/pull/366
- https://github.com/alphagov/collections-publisher/pull/247
- https://github.com/alphagov/publishing-api/pull/1010
 
https://trello.com/c/JF8IPgtL
https://trello.com/c/2AP3O9dn


README.md
```diff
@@ -14,6 +14,33 @@ And then execute:
...
+### Automatic error reporting
+
+If you include `govuk_app_config` in your `Gemfile` your application will be automatically configured to send errors to Sentry.
+
+### Manual error reporting
+
+Report something to Sentry manually:
+
+```rb
+GOVUK::Error.notify("Something went terrible")
```
whoojemaflip: `Something went terribly wrong`?

lib/govuk_app_config/error.rb
```diff
@@ -0,0 +1,15 @@
+require "sentry-raven"
+require "govuk_app_config/statsd"
+
+module GOVUK
+  module Error
```
whoojemaflip: This is a cool namespace, which makes me think it's possible there might be conflicts.

tijmenb: There aren't any conflicts, but on second thought, this is a inconsistent with what other gems provide. Perhaps we're better off with exposing `GovukError` & `GovukStatsd`.

README.md
```diff
@@ -14,6 +14,33 @@ And then execute:
 
 ## Usage
 
+### Automatic error reporting
+
+If you include `govuk_app_config` in your `Gemfile` your application will be automatically configured to send errors to Sentry.
```
boffbowsh: I think the magic should be documented somewhere. How does it do this?

nickcolley:
Out of curiosity, what's the reasoning behind the generic config direction rather than a shared govuk_sentry_config gem?

tijmenb:
@nickcolley Nothing gets past you! I'd like to consolidate the must-have configs/gems for GOV.UK app into this gem. That's probably unicorn, logstash, statsd & sentry. But this is a slightly longer term goal, and warrants some more discussion, so I didn't want to conflate this PR with that (but keep the option open with the name). I think Paul also has ideas about adding [Amazon X-Ray](https://aws.amazon.com/xray/) via this gem.

README.md
tijmenb: Added some more in https://github.com/alphagov/govuk_app_config/pull/1/commits/b3de5ab0bbd04e6cb9e3dcf2c81f69edf2a7d56a.

boffbowsh: Sorry, I meant things like using well known environment variables etc

tijmenb: That makes a lot of sense - https://github.com/alphagov/govuk_app_config/pull/1/commits/b1a3e28daf71435456ba19107c9de1bcf336f50e

PR #5, opened 2017-09-19
[WIP] Allow multiple `should_capture` filters for Sentry

alecgibson:
Sentry's Raven allows configuration of a method that determines
whether to capture errors or not: `should_capture=`. However, this
method only accepts a single `Proc` as an argument, and even then we
are already using this method in this gem in order to hook into Statsd.

In order to allow clients to use the `should_capture` method, this
change adds a `GovukError::Configuration` class, which sets up an array
of `Proc`s, which must _all_ return `true` to capture the error.

### DIFF
--- a/lib/govuk_app_config/configure.rb
+++ b/lib/govuk_app_config/configure.rb
@@ -6,9 +6,20 @@ Raven.configure do |config|
   # We need this until https://github.com/getsentry/raven-ruby/pull/736 is released
   config.current_environment = ENV["SENTRY_CURRENT_ENV"]
 
-  # We're misusing the `should_capture` block here to hook into raven until
-  # there's a better way: https://github.com/getsentry/raven-ruby/pull/750
-  config.should_capture = Proc.new {
+  config.should_capture = Proc.new do |error|
+    capture_filters = GovukError.configuration.capture_filters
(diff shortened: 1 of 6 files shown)


alecgibson:
[Only log an error in Errbit/Sentry when the Sidekiq job has no retries left](https://trello.com/c/5goqTLpz/550-only-log-an-error-in-errbit-sentry-when-the-sidekiq-job-has-no-retries-left)

tijmenb:
👍 this looks very useful!

lib/govuk_app_config/configure.rb
```diff
@@ -8,15 +8,23 @@
 
   # We're misusing the `should_capture` block here to hook into raven until
   # there's a better way: https://github.com/getsentry/raven-ruby/pull/750
-  config.should_capture = Proc.new {
-    GovukStatsd.increment("errors_occurred")
+  GovukError.configure do |error_config|
```
tijmenb: Does this need to be in the `Raven.configure` block?

alecgibson: Nope. Good point!

alecgibson:
We're going to solve the root cause of this PR using Sidekiq directly, rather than using Sentry. I'm closing this PR - it could be resurrected later if anyone has a need for this functionality.

PR #8, opened 2017-09-21
Add GovukError.configure

tijmenb:
This allows us to configure Raven without actually mentioning Raven in the code. Inspired by https://github.com/alphagov/govuk_app_config/pull/5 and [this comment](https://github.com/alphagov/content-performance-manager/pull/353/files/21e8ebfb275857387339625b68ba091fa1047c34#r140194431).

cc @alecgibson

### DIFF
--- a/lib/govuk_app_config/govuk_error.rb
+++ b/lib/govuk_app_config/govuk_error.rb
@@ -10,4 +10,10 @@ module GovukError
 
     Raven.capture_exception(exception_or_message, args)
   end
+
+  def self.configure
+    Raven.configure do |config|
+      yield(config)
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


lib/govuk_app_config/govuk_error.rb
```diff
@@ -10,4 +10,10 @@ def self.notify(exception_or_message, args = {})
 
     Raven.capture_exception(exception_or_message, args)
   end
+
+  def self.configure(&block)
```
alecgibson: Do you actually need `&block` here given that you're using `yield`?

tijmenb: We do not, thanks!

PR #10, opened 2017-10-10
Configure logging using this gem

tijmenb:
GOV.UK applications use logstasher to output logs as JSON so they can be shipped to Kibana. 

Over the years copy-and-changing has led to a complete circus of different logging configurations. 

Here's a list of applications and versions:

Application  | Logstasher version
-- | --
asset-manager | 0.4.8
authenticating-proxy | 0.6.5
calculators | 0.6.5
calendars | 1.2.1
collections | 0.6.2
collections-publisher | 1.2.1
content-performance-manager | 1.2.1
content-store | 0.5.0
content-tagger | 1.2.1
design-principles | 0.4.8
email-alert-api | 1.2.1
feedback | 0.6.2
finder-frontend | 0.4.9
frontend | 1.1.0
government-frontend | 0.6.1
hmrc-manuals-api | 0.5.3
imminence | 0.4.8
info-frontend | 0.6.0
licence-finder | 0.6.5
link-checker-api | 0.9.0
local-links-manager | 0.6.2
manuals-frontend | 0.6.2
manuals-publisher | 0.4.8
maslow | 0.4.8
publisher | 1.2.1
publishing-api | 0.6.2
release | 0.4.8
router-api | 0.6.2
service-manual-frontend | 0.6.1
service-manual-publisher | 1.2.1
short-url-manager | 0.5.3
signon | 0.4.8
smart-answers | 0.4.8
specialist-publisher | 0.6.2
static | 0.4.8
support | 0.4.8
support-api | 0.6.2
transition | 0.6.5
travel-advice-publisher | 0.4.8
whitehall | 1.2.1

And the following apps don't use logstasher at all:

- bouncer
- contacts-admin
- email-alert-frontend
- policy-publisher
- search-admin

This PR adds logstasher as a dependency and sets up an initialiser to set it up correctly. By centralising the config we can get rid of divergent config in different apps and use the same version everywhere.

We're doing this now because we need to [bump the logstasher gem in all apps to make it work with log.it](https://github.com/alphagov/govuk-puppet/pull/6650). Moving the config into this gem now will make future upgrade processes easier.

## Todo

- [x] ~Add tests~ it's crazy hard to test this..
- [x] Add documentation
- [x] Fix logstasher crashing: https://github.com/shadabahmed/logstasher/pull/142
- [x] Test with Rails app & non-Rails apps (bouncer, rummager, email-alert-service)

PR #11, opened 2017-10-10
Add Unicorn as a dependency

tijmenb:
All GOV.UK apps use unicorn as the web server, but we're using a wild variety of versions:

Name | Version
-- | --
asset-manager | 5.0.1
authenticating-proxy | 4.9.0
bouncer | 5.0.1
calculators | 5.0.1
calendars | 5.3.0
collections | 4.9.0
collections-publisher | 5.3.0
contacts-admin | 5.1.0
content-performance-manager | 5.3.0
content-store | 4.8.2
content-tagger | 5.0.1
design-principles | 4.3.1
email-alert-api | 5.3.0
email-alert-frontend | 5.3.0
feedback | 4.9.0
finder-frontend | 4.8.3
frontend | 4.9.0
government-frontend | 4.8.0
hmrc-manuals-api | 4.8.3
imminence | 4.3.1
info-frontend | 4.8.3
licence-finder | 5.1.0
link-checker-api | 5.3.0
local-links-manager | 4.9.0
manuals-frontend | 4.8.2
manuals-publisher | 4.8.2
maslow | 5.3.0
policy-publisher | 5.3.0
publisher | 5.3.0
publishing-api | 4.9.0
release | 4.4.0
router-api | 4.9.0
rummager | 5.1.0
search-admin | 4.9.0
service-manual-frontend | 4.8.0
service-manual-publisher | 5.3.0
short-url-manager | 5.1.0
signon | 4.9.0
smart-answers | 4.8.3
specialist-publisher | 4.9.0
static | 4.9.0
support | 4.9.0
support-api | 5.1.0
transition | 5.3.0
travel-advice-publisher | 5.0.1
whitehall | 5.3.0

This PR adds the gem as a dependency so we'll be using the same version. This will make debugging issues easier, as well as switching to a new web server like Puma.

Note that the email-alert-service isn't a web app so doesn't use a web server currently. I think it won't hurt making unicorn a dependency, for consistency's sake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+* Add Unicorn (our web server) as a dependency. Make sure to drop `gem 'unicorn'` from your Gemfile after upgrading.
+
 ## 0.3.0
 
 * Add `time` and `gauge` to `GovukStatsd`
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # GOV.UK Config
 
+Adds the basics of a GOV.UK application:
+
+- Unicorn as a web server
+- Error reporting with Sentry
+- Statsd client for reporting stats
+
 ## Installation
 
 Add this line to your application's Gemfile:
@@ -13,6 +19,12 @@ And then execute:
     $ bundle
 
 ## Usage
+
+## Unicorn
+
+No configuration required.
+
+## Error reporting
 
 ### Automatic error reporting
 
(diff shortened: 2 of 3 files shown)


tijmenb:
@kevindew the idea is that apps will remove the gem from their own Gemfile and let this gem manage the dependency. I've updated the changelog to be more explicit, does that make sense?

kevindew:
@tijmenb yup sure, just figured unless you're planning a big run one round (which seems like something you might do) we might have both for a while.

tijmenb:
@kevindew the Logit work is going to require us to bump all versions (https://github.com/alphagov/govuk-puppet/pull/6650, https://github.com/alphagov/govuk_app_config/pull/10) anyway, so I figured we could use the opportunity to do this as well.

PR #13, opened 2017-10-11
Revert "Bump Sentry client to 2.7.0"

tijmenb:
Reverts alphagov/govuk_app_config#12.

2.7.0 was yanked because of problems: https://github.com/getsentry/raven-ruby/issues/773


greysteil:
Sentry 2.7.1 has now been released with a fix for the issue in 2.7.0, and is running fine on all the codebases using Dependabot - you should be able to replace this PR with an update to 2.7.1 instead.

tijmenb:
Thanks @greysteil!

PR #15, opened 2017-11-06
Ignore `ActionController::BadRequest`

tijmenb:
This happens when a client makes an invalid request to the application. There's no need for us to act on these errors.

In terms of https://github.com/alphagov/govuk-rfcs/pull/87, this is a [Bad request error](https://github.com/alphagov/govuk-rfcs/blob/errors/rfc-087-dealing-with-errors.md#bad-request-errors) and should not be sent to Sentry. In future versions of the Sentry client it might even be ignored by default: https://github.com/getsentry/raven-ruby/pull/769.

Trello: https://trello.com/c/wX66HhtR

Examples:

- https://sentry.io/govuk/app-calculators/issues/354179330
- https://sentry.io/govuk/app-government-frontend/issues/364427882

### DIFF
--- a/lib/govuk_app_config/configure.rb
+++ b/lib/govuk_app_config/configure.rb
@@ -15,6 +15,18 @@ GovukError.configure do |config|
     true
   }
 
+  config.excluded_exceptions = [
+    'AbstractController::ActionNotFound',
+    'ActionController::InvalidAuthenticityToken',
+    'ActionController::RoutingError',
+    'ActionController::UnknownAction',
+    'ActiveJob::DeserializationError',
+    'ActiveRecord::RecordNotFound',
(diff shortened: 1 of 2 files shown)


lib/govuk_app_config/configure.rb
```diff
@@ -15,6 +15,19 @@
     true
   }
 
+  config.excluded_exceptions = [
```
thomasleese: If an application has its own `config.excluded_exceptions` set, will the two conflict in some way? Might it be worth doing `+=` here?

tijmenb: No, this just overwrites the raven-sentry's defaults (intentionally). If excluded exceptions is modified by the app it will happen after this code is run, so won't overwrite it.

PR #38, opened 2018-04-24
Add healthcheck support

elliotcm:
This is a port of alphagov/govuk_healthcheck@db03b4d

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,38 @@ GovukStatsd.gauge "bork", 100
 GovukStatsd.time("account.activate") { @account.activate! }
 ```
 
+## Healthchecks
+
+Set up a route in your rack-compatible Ruby application, and pick the built-in
+or custom checks you wish to perform.
+
+Custom checks must be any object which responds to `#call`.
+
+For Rails apps:
+```ruby
+module CustomCheck
+  def self.call
+    {
+      my_custom_check: {
+        status: ThingChecker.everything_okay? ? "OK" : "Critical"
+      }
+    }
+  end
+end
+```
+```ruby
+get "/healthcheck", to: GovukHealthcheck.rack_response(
+  GovukHealthcheck::SidekiqRedis,
+  GovukHealthcheck::ActiveRecord,
+  CustomCheck,
+)
+```
+
+This will check:
+- Redis connectivity (via Sidekiq)
+- Database connectivity (via ActiveRecord)
+- Your custom healthcheck
+
 ## Rails logging
 
 In Rails applications, the application will be configured to send JSON-formatted
--- a/README.md
+++ b/README.md
@@ -110,20 +110,33 @@ GovukStatsd.time("account.activate") { @account.activate! }
 Set up a route in your rack-compatible Ruby application, and pick the built-in
 or custom checks you wish to perform.
 
-Custom checks must be any object which responds to `#call`.
-
-For Rails apps:
+Custom checks must be any class or instance which implements
+[this interface](spec/lib/govuk_healthcheck/shared_interface.rb):
 ```ruby
 module CustomCheck
-  def self.call
+  def self.name
+    :custom_check
+  end
+
+  def self.status
+    ThingChecker.everything_okay? ? OK : CRITICAL
+  end
+
+  # Optional
+  def self.message
+    "This is an optional custom message"
+  end
+
+  # Optional
+  def self.details
     {
-      my_custom_check: {
-        status: ThingChecker.everything_okay? ? "OK" : "Critical"
-      }
+      extra: "This is an optional details hash",
     }
   end
 end
 ```
+
+For Rails apps:
 ```ruby
 get "/healthcheck", to: GovukHealthcheck.rack_response(
   GovukHealthcheck::SidekiqRedis,
--- a/README.md
+++ b/README.md
@@ -47,12 +47,6 @@ $ bundle exec unicorn -c config/unicorn.rb
 
 If you include `govuk_app_config` in your `Gemfile`, Rails' autoloading mechanism will make sure that your application is configured to send errors to Sentry.
 
-If you use the gem outside of Rails you'll have to explicitly require it:
-
-```rb
-require "govuk_app_config"
-```
-
 Your app will have to have the following environment variables set:
 
 - `SENTRY_DSN` - the [Data Source Name (DSN)][dsn] for Sentry
--- a/lib/govuk_app_config/govuk_healthcheck.rb
+++ b/lib/govuk_app_config/govuk_healthcheck.rb
@@ -9,6 +9,6 @@ module GovukHealthcheck
   end
 
   def self.healthcheck(checks)
-    Checkup.new(checks).as_json
+    Checkup.new(checks).run
   end
 end
--- a/lib/govuk_app_config/govuk_healthcheck.rb
+++ b/lib/govuk_app_config/govuk_healthcheck.rb
@@ -1,0 +1,14 @@
+require "govuk_app_config/govuk_healthcheck/checkup"
+require "govuk_app_config/govuk_healthcheck/active_record"
+require "govuk_app_config/govuk_healthcheck/sidekiq_redis"
+require "json"
+
+module GovukHealthcheck
+  def self.rack_response(*checks)
+    proc { [200, {}, [JSON.dump(healthcheck(checks))]] }
+  end
+
+  def self.healthcheck(checks)
+    Checkup.new(checks).as_json
+  end
+end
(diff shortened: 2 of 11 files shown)


lib/govuk_app_config/govuk_healthcheck.rb
```diff
@@ -0,0 +1,14 @@
+require "govuk_app_config/govuk_healthcheck/checkup"
+require "govuk_app_config/govuk_healthcheck/active_record" if defined?(ActiveRecord)
+require "govuk_app_config/govuk_healthcheck/sidekiq_redis"
```
tijmenb: Does this also need to be guarded with `if defined?(Sidekiq)`?

lib/govuk_app_config/govuk_healthcheck/checkup.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  class Checkup
+    # @param checks [Array] Array of objects/classes that respond to `run`
+    def initialize(checks)
+      @checks = checks
+    end
+
+    def as_json
+      {
+        status: 'ok',
```
tijmenb: Do you want to use the constant here too?

README.md
```diff
@@ -111,6 +111,40 @@ GovukStatsd.gauge "bork", 100
...
+    {
+      my_custom_check: {
+        status: ThingChecker.everything_okay? ? "OK" : "Critical"
+      }
+    }
+  end
+end
+```
+```ruby
+require "govuk_app_config/govuk_healthcheck"
```
tijmenb: Maybe add this to https://github.com/alphagov/govuk_app_config/blob/master/lib/govuk_app_config.rb so we don't have to `require`

README.md
```diff
@@ -111,6 +111,40 @@ GovukStatsd.gauge "bork", 100
...
+
+get "/healthcheck", to: GovukHealthcheck.rack_response(
```
tijmenb: Would it be worth it to set up an engine so we can do something like this?

```rb
mount GovukAppConfig::Engine, at: "/healthcheck"
```

lib/govuk_app_config/govuk_healthcheck.rb
elliotcm: I'm not certain either are needed tbh, I ran info-frontend with this code and it doesn't have either sidekiq or AR

README.md
elliotcm: Maybe, though perhaps not in the first pass.  Right now there's not really much to wrap.

lib/govuk_app_config/govuk_healthcheck.rb
elliotcm: Removed from both.

lib/govuk_app_config/govuk_healthcheck/checkup.rb
elliotcm: Done

README.md
elliotcm: Done

lib/govuk_app_config/govuk_healthcheck/checkup.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  class Checkup
+    # @param checks [Array] Array of objects/classes that respond to `run`
+    def initialize(checks)
+      @checks = checks
+    end
+
+    def as_json
+      {
+        status: OK,
```
tijmenb: https://github.com/alphagov/email-alert-api/pull/234 seems to imply that the overall status is important too, do you know what the difference is to Icinga?

thomasleese:
I'm okay with this kind of common health check library, but I wonder if we should instead make this more similar to the health check system we put in `email-alert-api`:

- [`DatabaseHealthcheck`](https://github.com/alphagov/email-alert-api/blob/master/app/models/healthcheck/database_healthcheck.rb)
- [`Healthcheck`](https://github.com/alphagov/email-alert-api/blob/master/app/models/healthcheck.rb)
- [`HealthcheckController`](https://github.com/alphagov/email-alert-api/blob/master/app/controllers/healthcheck_controller.rb)

The API for creating individual health checks is more restrictive than returning a hash, but it means that it's less likely to accidentally make a mistake in the health check JSON format and makes them a lot easier to test.

Similarly, we can [add support for the `message` key which gets shown in Icinga](https://github.com/alphagov/govuk-puppet/blob/88ce80007e1151de15aba231b09bfbbd136aed99/modules/icinga/files/usr/lib/nagios/plugins/check_json_healthcheck#L134) which is [currently used in Whitehall](https://github.com/alphagov/whitehall/blob/master/app/controllers/healthcheck_controller.rb#L25).

elliotcm:
@thomasleese @tijmenb @bevanloon Thomas's suggests seem fine but they will change the behaviour of the original healthcheck gem which was intended to allow simple objects (or procs) to act as these healthchecks.

Are we all happy with that?

bevanloon:
Sounds like a reasonable suggestion to me - if the result are easier to test and harder to break accidentally, lets go for that.

elliotcm:
@thomasleese Updated.

README.md
```diff
@@ -111,6 +105,38 @@ GovukStatsd.gauge "bork", 100
...
+## Healthchecks
+
+Set up a route in your rack-compatible Ruby application, and pick the built-in
+or custom checks you wish to perform.
+
+Custom checks must be any object which responds to `#call`.
+
+For Rails apps:
+```ruby
+module CustomCheck
```
thomasleese: Is this documentation still correct?

lib/govuk_app_config/govuk_healthcheck/checkup.rb
```diff
@@ -0,0 +1,46 @@
...
+        status: worst_status,
+        checks: component_statuses,
+      }
+    end
+
+  private
+
+    def component_statuses
+      @component_statuses ||= @checks.each_with_object({}) do |check, hash|
+        check_result = check.details.merge(status: check.status)
```
thomasleese: Could we make `check.details` optional, like `check.message`?

spec/lib/govuk_healthcheck/active_record_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "spec_helper"
+require "govuk_app_config/govuk_healthcheck"
+require_relative "shared_interface"
+
+RSpec.describe GovukHealthcheck::ActiveRecord do
+  module ActiveRecord
```
thomasleese: Do we need this if we have the `allow(...)` on `ActiveRecord::Base` below?

lib/govuk_app_config/govuk_healthcheck/checkup.rb
```diff
@@ -0,0 +1,46 @@
+module GovukHealthcheck
+  STATUSES = [
```
thomasleese: I don't see a benefit to just returning `:ok:`, `:warning` and `:critical` from the `status` method here, since they're the same name, and we can have the `it_behaves_like "a healthcheck"` test check that they are valid like it does now.

README.md
elliotcm: NO

lib/govuk_app_config/govuk_healthcheck/checkup.rb
elliotcm: Sure

lib/govuk_app_config/govuk_healthcheck/checkup.rb
elliotcm: Well, bearing in mind I didn't write that originally so I don't know the reasons for it being there, I would say that we can't use the shared interface tests for that because they don't know enough about each class to trigger it to produce each status.

thomasleese: Hmm, I suppose that's true. I don't mind too much about this, but it doesn't feel very Ruby.

README.md
elliotcm: Updated.

lib/govuk_app_config/govuk_healthcheck/checkup.rb
elliotcm: Updated.

spec/lib/govuk_healthcheck/active_record_spec.rb
elliotcm: This has been changed to a better approach now.

elliotcm: (but the answer is "yes" because you can't use `allow` to stub an unimplemented method)

elliotcm:
@thomasleese Probably best to update the CHANGELOG as part of the version bump since it's good practice to claim a version number at the point of publish to avoid clashes, but thanks for the reminder.

thomasleese:
@elliotcm I prefer to update the `CHANGELOG` in the PR that made the change (even if it's under a `Unreleased` version heading) in case a few people have worked on features for a release and the person who puts together the version bump together won't risk forgetting anything.

elliotcm:
@thomasleese Fair enough.

PR #45, opened 2018-06-28
Expect health checks to be classes not instances

thomasleese:
While adding support for more advanced health checks in email-alert-api we switched over to using this framework provided by govuk_app_config.

One thing we've discovered is that by having health checks be a long running class instance, we can't cache any queries and a single health check could end up running the same query three times: for the status, details and message.

We could change the way health checks work and make them more complicated in email-alert-api, but I suspect this will be a problem which affects many other apps so I decided we should change the behaviour of govuk_app_config to expect health checks to be classes, which can then be instanced at the time of the request. This also follows the same behaviour of email-alert-api before it switched over to using this interface.

This is a breaking change, but since only email-alert-api uses custom checks at the moment and the interface hasn't changed for the built in checks, I think it's fine to consider it only a major update and the next release will be 1.6.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Make health checks classes rather than instances, allowing internal data to
+  be cached and improve performance.
+
 # 1.5.1
 
 * Set the `Content-Type` of healthchecks to `application/json`.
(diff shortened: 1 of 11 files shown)


barrucadu:
I like the change, I don't like the version number so much.

thomasleese:
@barrucadu happy to bump to 2.0.0 instead, just wanted to avoid a big version bump for something which until a lot of our apps use I would still consider in development/prototype.

steventux:
I think it's a sensible change. :metal:

PR #50, opened 2018-08-05
Stop saying errors are not going to be captured in dev

cbaines:
This is useful to have logged in production, but is distracting
everywhere else.


benthorner:
Loving this change 😃.

PR #52, opened 2018-08-30
Add GovukNodes

elliotcm:
This provides functionality for fetching all DNS names for a given node class, regardless of whether on AWS or Carrenza.

Required for clearing varnish cache, since the `backend` machines need to issue calls directly to each `cache` box.

https://trello.com/c/f8U7EfUh/427-make-the-new-cache-clearing-app-invalidate-cache-in-varnish

### DIFF
--- a/lib/govuk_app_config/govuk_nodes/carrenza_fetcher.rb
+++ b/lib/govuk_app_config/govuk_nodes/carrenza_fetcher.rb
@@ -11,7 +11,7 @@ class GovukNodes
   private
 
     def instances_of_class(node_class)
-      url = "http://puppetdb.cluster/v2/nodes?#{query_string(node_class)}"
+      url = "#{puppetdb_node_url}?#{query_string(node_class)}"
       json_response = open(url).read
       JSON.parse(json_response)
     end
@@ -21,6 +21,10 @@ class GovukNodes
       query = %{["or", ["~", ["fact", "fqdn"], "^#{hyphenated_node_class}-\\d+."]]}
 
       URI.encode_www_form(query: query)
+    end
+
+    def puppetdb_node_url
+      @puppetdb_node_url ||= ENV["PUPPETDB_NODE_URL"]
     end
   end
 end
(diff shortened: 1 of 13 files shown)


lib/govuk_app_config/govuk_nodes/carrenza_fetcher.rb
```diff
@@ -0,0 +1,26 @@
...
+    def hostnames_of_class(node_class)
+      instances_of_class(node_class).map do |instance|
+        instance.fetch("name")
+      end
+    end
+
+  private
+
+    def instances_of_class(node_class)
+      url = "http://puppetdb.cluster/v2/nodes?#{query_string(node_class)}"
```
deanwilson: Can this be provided by a env.d variable or some such? Hard coded names make me nervous.

elliotcm: It's hardcoded in the script I ported, but sure.  Do you have somewhere I can source it from in puppet, or should I hardcode it there?

elliotcm: Or I can put it in hiera but I'm assuming it's somewhere already

elliotcm: @deanwilson Updated.

tijmenb:
I don't have any context on the project, but I can't figure out where this code will be used. Would it be possible to add some context to the README/docs about the purpose and intended usage?

elliotcm:
@tijmenb sure

elliotcm:
@tijmenb Updated.

tijmenb:
@elliotcm thanks! If this will only be used by the cache clearing service, would it be better to put the code there initially? It would be good to keep govuk_app_config for app configuration, this seems more of a shared utility.

elliotcm:
I had a feeling you were going to suggest that.  No problem, can do.

elliotcm:
Incorporated into https://github.com/alphagov/cache-clearing-service/pull/6

PR #53, opened 2018-09-11
Enable AWS X-Ray tracing

barrucadu:
This is dependent on the puppet stuff being done.

---

[Trello card](https://trello.com/c/PFXPhBxO/451-install-default-x-ray-gem-for-rails-in-all-our-apps-mandatory-%F0%9F%8D%90)


lib/govuk_app_config/configure.rb
```diff
@@ -42,3 +42,14 @@
     GovukStatsd.increment("error_reports_failed")
   }
 end
+
+Rails.application.config.xray = {
```
thomasleese: Might be nice to extract this into an `xray.rb` file and then include it.

lib/govuk_app_config/configure.rb
```diff
@@ -42,3 +42,14 @@
...
+  name: ENV["GOVUK_APP_NAME"].to_s,
+  patch: %I[net_http aws_sdk],
+  sampling_rules: {
+    default: {
+      "fixed_target": 0,
+      "rate": 0.01,
```
thomasleese: Might be nice to make this configurable, but with a default of `0.01`.

lib/govuk_app_config/configure.rb
barrucadu: Done

lib/govuk_app_config/configure.rb
barrucadu: Done

lib/govuk_app_config/xray.rb
```diff
@@ -0,0 +1,12 @@
+require 'aws-xray-sdk/facets/rails/railtie'
+
+Rails.application.config.xray = {
+  name: ENV["GOVUK_APP_NAME"].to_s,
+  patch: %I[net_http aws_sdk],
+  sampling_rules: {
+    default: {
+      "fixed_target": ENV.fetch("GOVUK_XRAY_SAMPLE_TARGET", 0).to_i,
```
thomasleese: I don't think we need the `GOVUK_` prefix on these, and I wonder if we'll need to cast them to floats before setting the argument?

barrucadu: Casting beyond the current `.to_i` and `.to_f`?

thomasleese: Not sure how I missed that...

lib/govuk_app_config/govuk_xray.rb
```diff
@@ -0,0 +1,28 @@
+require 'aws-xray-sdk/facets/rack'
+require 'aws-xray-sdk/facets/rails/ex_middleware'
+
+module GovukXRay
+  def self.initialize(app)
+    app.middleware.insert 0, XRay::Rack::Middleware
+    app.middleware.use XRay::Rails::ExceptionMiddleware
+  end
+
+  def self.start
```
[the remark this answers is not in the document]
barrucadu: I went for a different name because it doesn't just set configuration, it also starts the recorder. 
 And unlike `GovukLogging` it's called in the `after_initialize` bit of the railtie.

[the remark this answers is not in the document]
thomasleese: That's true, I don't mind either way.

PR #57, opened 2018-09-25
Set a fallback XRay recorder name if GOVUK_APP_NAME is missing

barrucadu:


lib/govuk_app_config/govuk_xray.rb
```diff
@@ -12,8 +12,13 @@ def self.start
     patch = Gem.loaded_specs.has_key?('aws-sdk-core') ?
               %I[aws_sdk net_http] : %I[net_http]
 
+    # if there isn't a name set, attempting to record a segment will
+    # throw an error
+    govuk_app_name = ENV.fetch('GOVUK_APP_NAME')
```
thomasleese: I think you'll want to use `ENV.fetch("GOVUK_APP_NAME", "xray")` to avoid an exception being raised.

barrucadu: Whoops, that should have been `ENV['GOVUK_APP_NAME']`.  The reason I don't just do `ENV.fetch('GOVUK_APP_NAME', 'xray')` is because that won't catch the environment variable being present but empty.

PR #61, opened 2018-10-03
Check for XRAY_PATCH_AWS_SDK env var, and do not patch aws_sdk if unset

barrucadu:
**I'll open an issue on aws-xray-sdk if other people agree with my conclusion.**

---

Since applying the last round of govuk_app_config upgrades, asset-manager and publishing-api have been using far more memory:

asset-manager
publishing-api

The memory shoots up as soon as the app is deployed, so I believe this is an overhead imposed at initialisation-time, rather than some sort of leak which manifests when requests are made.

After doing some debugging (commenting out lines of code and crossing my fingers), I narrowed this down to having X-Ray patch the `aws_sdk` gem.  If that gem is patched, the problem occurs; if it is not, even if `net_http` is patched, the memory usage seems much closer to before.

It's a shame to lose tracking for stuff which goes through `aws_sdk`, but that's only a small amount of stuff (asset uploads from the asset-manger, and nothing(!) from the publishing-api), and I think we can assume that AWS is not the bottleneck in our architecture.

---

[Trello card](https://trello.com/c/uaiWiI51/515-investigate-increased-memory-use-of-asset-manager-and-publishing-api-since-x-ray-deploy)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+* Only instrument the `aws_sdk` gem with AWS X-Ray if the
+  `XRAY_PATCH_AWS_SDK` environment variable is present.
+
 # 1.9.3
 
 * Do not report Sidekiq queue thresholds in healthchecks which are
--- a/lib/govuk_app_config/govuk_xray.rb
+++ b/lib/govuk_app_config/govuk_xray.rb
@@ -8,9 +8,10 @@ module GovukXRay
   end
 
   def self.start
-    # if aws-sdk is loaded, we want to instrument that too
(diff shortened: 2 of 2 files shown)


barrucadu:
Here are some more graphs.

asset-manager continues to look good, memory usage has been creeping up but it's not yet at the point it was before X-Ray, and it's quite slow:

asset-manager

publishing-api was good, then a branch with X-Ray instrumenting `aws_sdk` was deployed.  The big spike will be due to the old processes and new processes temporarily co-existing (this doesn't happen in the earlier graphs, as I deployed by killing the old processes first).  Memory usage is below what it was, but is still rather higher than it was without X-Ray instrumenting `aws_sdk`:

publishing-api

PR #63, opened 2019-01-03
Allow disabling X-Ray railtie through env var

barrucadu:
The setting up of dummy classes for the test code is a bit unpleasant.


barrucadu:
There's no use-case for disabling logging that I can think of, but I was thinking about disabling railties in general.  But preventing that concept from leaking would be nicer.

PR #69, opened 2019-01-10
Logging improvements

cbaines:


lib/govuk_app_config/govuk_logging.rb
```diff
@@ -20,7 +20,7 @@ def self.configure
     $stdout.reopen($stderr)
 
     # Send Rails' logs to STDERR because they're not JSON formatted.
-    Rails.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr, level: Logger::INFO))
+    Rails.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr, level: Rails.logger.level))
```
pmanrubia: Would it be possible to introduce a high logging rate if the application has a `debug` logger level assigned by default? It seems that we are overwriting it.

cbaines: Currently, I think the logging level from the application is being change here when `Rails.logger` is replaced. I'm not sure I quite follow your question, but I think the answer is yes.

This change may result in the log level changing in applications, if the application is setting the log level to `debug`, then it'll be kept as `debug` here.

PR #70, opened 2019-01-22
Upgrade Sentry client and use new features

tijmenb:
See commits for details.

### DIFF
--- a/lib/govuk_app_config/configure.rb
+++ b/lib/govuk_app_config/configure.rb
@@ -10,16 +10,6 @@ GovukError.configure do |config|
     GovukStatsd.increment("errbit.errors_occurred")
   }
 
-  config.should_capture = Proc.new { |e|
-    exception_class = e.respond_to?(:original_exception) ? e.original_exception.class : e.class
-    if exception_class.ancestors.any? { |c| c.name =~ /^GdsApi::(HTTPIntermittent|TimedOutException)/ }
-      GovukStatsd.increment("gds_api_adapters.errors.#{e.class.name.demodulize.underscore}")
-      false
-    else
-      true
-    end
-  }
-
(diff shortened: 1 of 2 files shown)


lib/govuk_app_config/configure.rb
```diff
@@ -3,21 +3,11 @@
...
-  config.should_capture = Proc.new { |e|
+  config.before_send = Proc.new { |e|
     GovukStatsd.increment("errors_occurred")
 
     # For backwards compatibility
     GovukStatsd.increment("errbit.errors_occurred")
-
-    exception_class = e.respond_to?(:original_exception) ? e.original_exception.class : e.class
-    if exception_class.ancestors.any? { |c| c.name =~ /^GdsApi::(HTTPIntermittent|TimedOutException)/ }
-      GovukStatsd.increment("gds_api_adapters.errors.#{e.class.name.demodulize.underscore}")
```
thomasleese: I wonder if there's any way we can keep this behaviour? It was useful being able to see graphs like this:

download

Perhaps we could even make it generic and just record stats for every kind of error?

tijmenb: Would `GovukStatsd.increment("errors.#{e.class.name.demodulize.underscore}")` work? It would be generic but not backwards compatible.

thomasleese: I think that sounds good - we can always combine the old and the new data in graphs if you need both.

kevindew: +1 for this - on Content Publisher we're trying to track how a big a problem the HTTP intermittents are for us

tijmenb:
I'm going to close this until I have more time to finish it (feel free to pick this up in the meantime).

PR #71, opened 2019-02-07
Enable initialized health check classes

pmanrubia:
Adds support for adding parametrized classes for the health check chain:

```ruby
get "/healthcheck", to: GovukHealthcheck.rack_response(
  GovukHealthcheck::SidekiqRedis,
  GovukHealthcheck::ActiveRecord,
  GovukHealthcheck::InitializedCheck.new(:param),
)
```

Enable applications to reuse the same Check without being forced
to use inheritance or other mechanisms.

This is introduced in https://github.com/alphagov/content-performance-manager/pull/1131/commits/73aec09c289f785c06040845662c61a7f4d163e1, and is part of PR https://github.com/alphagov/content-performance-manager/pull/1131. We aim to reuse the same check for multiple metrics.

### DIFF
--- a/docs/healthchecks.md
+++ b/docs/healthchecks.md
@@ -58,6 +58,14 @@ get "/healthcheck", to: GovukHealthcheck.rack_response(
 )
 ```
 
+It also accepts objects, so classes can be initialized:
+
+```ruby
+get "/healthcheck", to: GovukHealthcheck.rack_response(
+  InitializedCheck.new(:param),
+)
+```
+
 ## Built-in Checks
 
 A convention used when naming these classes is that it should end with `Check`
--- a/spec/lib/govuk_healthcheck/checkup_spec.rb
+++ b/spec/lib/govuk_healthcheck/checkup_spec.rb
@@ -63,9 +63,18 @@ RSpec.describe GovukHealthcheck::Checkup do
     expect(response.dig(:checks, :critical_check, :status)).to eq(:ok)
   end
 
+  it "accepts objects (can be initialized)" do
+    response = described_class.new([TestHealthcheck.new]).run
+    expect(response.dig(:checks, :unknown_check, :status)).to eq(:unknown)
+  end
+
   class TestHealthcheck
     def name
       "#{status}_check".to_sym
+    end
+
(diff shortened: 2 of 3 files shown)


pmanrubia:
@tijmenb thank you for your feedback. I have addressed your comments. Could you please have another look?

spec/lib/govuk_healthcheck/checkup_spec.rb
```diff
@@ -63,6 +64,11 @@
     expect(response.dig(:checks, :critical_check, :status)).to eq(:ok)
   end
 
+  it "accepts Objects (can be initialized)" do
+    response = described_class.new([InitializedCheck.new(:param)]).run
```
tijmenb:
```suggestion
    response = described_class.new([TestHealthcheck.new]).run
```

You could use `TestHealthcheck.new` here instead, it'll test the same thing and we don't have to define the special `InitializedCheck` class.

docs/healthchecks.md
```diff
@@ -58,6 +58,14 @@ get "/healthcheck", to: GovukHealthcheck.rack_response(
 )
 ```
 
+It also accepts Objects, so classes can be initialized:
```
tijmenb:
```suggestion
It also accepts objects, so classes can be initialised:
```

spec/lib/govuk_healthcheck/checkup_spec.rb
```diff
@@ -7,6 +7,7 @@
   let(:critical_check) { CriticalTestHealthcheck }
   let(:disabled_critical_check) { DisabledCriticalHealthcheck }
   let(:exception_check) { ExceptionHealthcheck }
+  let(:parametrized_check) { InitializedCheck.new(:param) }
```
tijmenb: I think this can be removed:

```suggestion
```

spec/lib/govuk_healthcheck/checkup_spec.rb
pmanrubia: I have removed it, although I see value in setting a fixture that is close to what we are testing (or how it is used)

PR #89, opened 2019-05-22
Change context_missing error level

bilbof:
This will remove persistent logging of `ERROR -- : can not find the current context.` from our logs.

This is from the aws-xray-sdk-ruby gem. Changing the log level in this way seems to have no side effects.

See: https://github.com/aws/aws-xray-sdk-ruby/blob/5134f203a0a0c7681e137aa6045ad64fc645c8fe/lib/aws-xray-sdk/context/default_context.rb#L62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Don't log Context Missing Errors (`ERROR -- : can not find the current context.`)
+
 # 1.16.1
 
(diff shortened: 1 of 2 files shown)


bilbof:
@sihugh added a note to the changelog

kevindew:
@bilbof Turns out this isn't released yet in aws-xray-sdk gem

PR #90, opened 2019-05-22
Tweak our CSP to work with 'dev.gov.uk'

benthorner:
https://trello.com/c/j7mUrNuO/883-fix-csp-breaking-apps-in-development-environment

    Previously our Content Security Policy (CSP) disallowed content from
    our own apps when run in development with the 'dev.gov.uk' domain. Our
    frontend apps rely on externals assets (e.g. from static.dev.gov.uk)
    and these are not allowed under the current policy.
    
    We also hardcode some assets, such as 'placeholder.png' to be served
    from the production assets domain. This means that we should allow for
    both 'dev.gov.uk' and 'publishing.service.gov.uk' in development. But we
    should avoid implicitly trusting 'dev.gov.uk' outside of development.
    
    This also removes 'localhost' from all policies, as it should normally
    be covered by 'self', which applies to content requested from the same
    domain and port. It would impact running connecting multiple apps on
    localhost with different ports, but there's no clear need to do this.

Screen Shot 2019-05-22 at 13 30 44

### DIFF
--- a/lib/govuk_app_config/govuk_content_security_policy.rb
+++ b/lib/govuk_app_config/govuk_content_security_policy.rb
@@ -11,7 +11,11 @@ module GovukContentSecurityPolicy
   # - https://csp-evaluator.withgoogle.com
   # - https://cspvalidator.org
 
-  GOVUK_DOMAINS = "'self' *.publishing.service.gov.uk localhost".freeze
+  GOVUK_DOMAINS = [
+    "'self'",
+    '*.publishing.service.gov.uk',
+    "*.#{ENV['GOVUK_APP_DOMAIN_EXTERNAL'] || ENV['GOVUK_APP_DOMAIN'] || 'dev.gov.uk'}"
+  ].uniq.join(" ").freeze
 
   GOOGLE_ANALYTICS_DOMAINS = "www.google-analytics.com ssl.google-analytics.com stats.g.doubleclick.net".freeze
 
@@ -130,9 +134,9 @@ module GovukContentSecurityPolicy
     # AWS Lambda function that filters out junk reports.
     if Rails.env.production?
       reporting = "report-uri https://jhpno0hk6b.execute-api.eu-west-2.amazonaws.com/production"
-      Rails.application.config.action_dispatch.default_headers['Content-Security-Policy-Report-Only'] = GovukContentSecurityPolicy.build + " " + reporting
+      Rails.application.config.action_dispatch.default_headers['Content-Security-Policy-Report-Only'] = self.build + " " + reporting
     else
-      Rails.application.config.action_dispatch.default_headers['Content-Security-Policy'] = GovukContentSecurityPolicy.build
+      Rails.application.config.action_dispatch.default_headers['Content-Security-Policy'] = self.build
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Tweak our CSP to work with 'dev.gov.uk'
+
 # 1.16.3
 
 * Revert PR #89 - it relies on an unreleased feature of aws-xray-sdk


benthorner:
(Alternatively, to better achieve parity between dev and production, we might want to consider moving to something like 'dev.publishing.service.gov.uk', but then that's more typing :-().

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -11,7 +11,8 @@ module GovukContentSecurityPolicy
   # - https://csp-evaluator.withgoogle.com
   # - https://cspvalidator.org
 
-  GOVUK_DOMAINS = "'self' *.publishing.service.gov.uk localhost".freeze
+  PROD_GOVUK_DOMAINS = "'self' *.publishing.service.gov.uk localhost".freeze
```
erino: Now that `GOVUK_DOMAINS` is broken into dev and prod. Is `localhost` needed in production?

benthorner: Good point. I can't think of a realistic scenario for running in prod on localhost (i.e. dev machine), since a bunch of other stuff usually breaks when you try that anyway.

benthorner: I've added a commit to remove it in prod.

erino: @rubenarakelyan is there reason for having `localhost` in production?

kevindew: I was asking Ruben about that yesterday. I think it should be fine to go as I think it expects to connect to localhost over https and I doubt that happens. Ruben thought it might be needed as part of tests but wasn't sure.

So worth having a go at killing it, may try running this branch against government-frontend if you want to be extra cautious?

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -124,15 +125,19 @@ def self.build
     policies.map { |str| str.join(" ") }.join("; ") + ";"
   end
 
+  def self.govuk_domains
+    Rails.env.production? ? PROD_GOVUK_DOMAINS : DEV_GOVUK_DOMAINS
+  end
```
kevindew: I don't think we should use the env as an indicator of what domains we'd expect. Rails production mode shouldn't be exclusively for running on production servers and it is used as the environment for running end-to-end tests (which is a reason why we should be doing https://github.com/alphagov/govuk_app_config/blob/243eaca4a23a1e3c4fb835d81f668e3c533b49e3/lib/govuk_app_config/govuk_content_security_policy.rb#L132 via an env var rather than hardcoding).

I also think we'll need `*.publishing.service.gov.uk` in all environments (e.g accessing assets or using [live static](https://github.com/alphagov/government-frontend/blob/83ab8d4fe9df05753b6821d16dde85ce3c447ae5/startup.sh#L15)) 

How much of a concern do you think it is to just allow `*.dev.gov.uk` globally? Some of our options around this are using Plek or the GOVUK_APP_DOMAIN env vars. Hopefully there aren't real needs for localhost in either dev or prod.

benthorner:
@kevindew @erino I've squashed a pushed a change that hopefully reflects your suggestions. This commit explains in a little more detail, but in summary:

   - We do indeed need *.publishing.... in development, although it seems rather arbitrary
   - We shouldn't need localhost as it's usually covered by 'self'
   - We should avoid the surprise of allowing dev.gov.uk in production, so I've done a compromise

Unfortunately Plek isn't included in govuk_app_config, so I've nicked a bit of it's implementation around domain environment variables, which isn't ideal - what do you think about depending on it?

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -11,7 +11,11 @@ module GovukContentSecurityPolicy
   # - https://csp-evaluator.withgoogle.com
   # - https://cspvalidator.org
 
-  GOVUK_DOMAINS = "'self' *.publishing.service.gov.uk localhost".freeze
+  GOVUK_DOMAINS = [
+    "'self'",
+    '*.publishing.service.gov.uk',
+    "*.#{ENV['GOVUK_APP_DOMAIN'] || ENV['DEV_DOMAIN'] || 'dev.gov.uk'}"
```
kevindew: Nice idea 👍 

I know this will probably make you sad but we should probably start with GOVUK_APP_DOMAIN_EXTERNAL before GOVUK_APP_DOMAIN - since otherwise this may end up with the .digital internal AWS domains. 

Have you seen DEV_DOMAIN used outside plek at all? Since it's a non GOVUK env var I wonder if we can drop it here - feels like you should just use GOVUK_APP_DOMAIN for it anyway.

benthorner: Good point :-). I kind of wish we'd lead with 'DOMAIN_INTERNAL' on the other one, since that's what they sometimes are. Anyway, I've updated as suggested, since I agree 'DEV_DOMAIN' should use an official prefix if it wants to be supported.

benthorner:
A companion PR to completely stop supporting ENV['DEV_DOMAIN']: https://github.com/alphagov/plek/pull/64

lib/govuk_app_config/govuk_content_security_policy.rb
kevindew: I agree about domain internal, it's a pain point. Sadly you'll have to also check GOVUK_APP_DOMAIN after GOVUK_APP_DOMAIN_EXTERNAL as the external one is not always set :(

Have you been able to test government-frontend to check it's tests still pass with these changes? 

Nice one with regards to Plek 👍

benthorner: 😭

benthorner: Forgot to mention - I've checked government-frontend after each change. All I can say is that it works for dev.gov.uk, and the fact that we still hardcode publishing... means the functionality won't change otherwise.

kevindew:
Btw this should have a changelog update.

PR #92, opened 2019-05-24
Use Rails DSL for CSP

kevindew:
This is built off https://github.com/alphagov/govuk_app_config/pull/90 with the expectation that will be merged first.

This refactors the CSP to use the Rails DSL, this is backwards
compatible with the previous implementation. This refactor provides a few
benefits:

- no longer needs to set up the actual policy, just configuration
  needed;
- apps can add/remove things to the policy if needed;
- the nonce generator allows setting a nonce on inline javascript tags

I've reduced the amount of comments in this file for brevity. A lot of
the example URLs felt unnecessary for the source code and felt more like they
belonged in a commit message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Use Rails DSL to configure content security policy, allowing apps to modify
+  the policy and use nonce features.
+
 # 1.17.0
 
 * Tweak our CSP to work with 'dev.gov.uk'
--- a/govuk_app_config.gemspec
+++ b/govuk_app_config.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "unicorn", "~> 5.4.0"
 
   spec.add_development_dependency "bundler", "~> 1.15"
+  spec.add_development_dependency "rails", "~> 5"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.6.0"
   spec.add_development_dependency "rspec-its", "~> 1.2.0"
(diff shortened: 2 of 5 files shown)


benthorner:
It looks like we've lost the setting for the report_uri, or whatever the equivalent is in DSL land. I've also had a go at using this with government-frontend in development, and encountered one error:

```
test-44554:69 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' *.publishing.service.gov.uk *.dev.gov.uk www.google-analytics.com ssl.google-analytics.com stats.g.doubleclick.net www.signin.service.gov.uk *.ytimg.com www.youtube.com 'unsafe-inline' 'nonce-OX35waDJswFurIuCAB+6QA=='". Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list.
```

Since the nonce appears to be new, could it be that this introduces a bug?

kevindew:
> It looks like we've lost the setting for the report_uri, or whatever the equivalent is in DSL land. 

Ah shoot! that was in an earlier version. I'll bring that back.

> I've also had a go at using this with government-frontend in development, and encountered one error:
> 
> ```
> test-44554:69 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' *.publishing.service.gov.uk *.dev.gov.uk www.google-analytics.com ssl.google-analytics.com stats.g.doubleclick.net www.signin.service.gov.uk *.ytimg.com www.youtube.com 'unsafe-inline' 'nonce-OX35waDJswFurIuCAB+6QA=='". Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list.
> ```
> 
> Since the nonce appears to be new, could it be that this introduces a bug?

Oh that's a bit annoying, I wasn't aware of that I'll have to pull that out and have apps configure the nonce inline 😞. Thanks for catching both of those issues.

kevindew:
@benthorner I've fixed both of those issues, and successfully ran the government-frontend test suite with it

PR #99, opened 2019-07-08
Add youtube-nocookie.com to CSP

vanitabarrett:
## What
Adds `youtube-nocookie.com` to the GOVUK Content Security Policy.

## Why
We are [updating the media player on GOV.UK to use the Youtube player](https://github.com/alphagov/govuk_publishing_components/pull/908) rather than Nomensa accessible player. This change involves using Youtube's Privacy Enhanced mode (youtube-nocookie.com). Our current security policy only allows `youtube.com`, so we need to add `youtube-nocookie.com`

### DIFF
--- a/lib/govuk_app_config/govuk_content_security_policy.rb
+++ b/lib/govuk_app_config/govuk_content_security_policy.rb
@@ -38,6 +38,7 @@ module GovukContentSecurityPolicy
                       # Allow YouTube Embeds (Govspeak turns YouTube links into embeds)
                       "*.ytimg.com",
                       "www.youtube.com",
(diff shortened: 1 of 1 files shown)


lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -38,6 +38,7 @@ def self.build_policy(policy)
                       # Allow YouTube Embeds (Govspeak turns YouTube links into embeds)
                       "*.ytimg.com",
                       "www.youtube.com",
+                      "www.youtube-nocookie.com"
```
kevindew: Looks like you're missing a comma here.

PR #113, opened 2019-07-15
Send Rails.logger calls to stdout as JSON

kevindew:
Trello: https://trello.com/c/P3z1xFBW/968-investigate-methods-to-prevent-access-to-document-based-on-access-limit-and-state

This changes the approach to logging so that calls to Rails.logger are
included in the stdout of the app and are included in Kibana. Previously
they were sent to stdout. This brings the logging to
be consistent with the approach that is used for sidekiq and with the
comment at the top of govuk_logging file.

This works by creating a new formatter that creates a JSON line that is
output to stdout.

It has a couple of things that make it not perfect:

1) When using with `rails server` log output is outputted twice. Rails
has special code to stop duplicate output to stdout [1], however the
redirect we make with $stdout to $stderr breaks this. This PR doesn't
introduce this problem, but was something I witnessed in testing.
2) Tagged logging updates the message and can't be used in JSON payload.
When testing this with Content Publisher I noticed it was logging a
@message of "[1c52e5f4-6e1d-41dc-a1cd-90a00a468f54] testing logging",
where the UUID is a request_id. I hoped that this is something that
could be a different JSON field however the Rails tagged logging only
has values and not keys [2]

[1]: https://github.com/rails/rails/commit/3d10d9d6c3b831fe9632c43a0ffec46104f912a7
[2]: https://github.com/rails/rails/blob/bf19b8774e20e98f7fdcd3ac82ee17f9adee22d8/railties/lib/rails/rack/logger.rb#L26


kevindew:
> This looks good to me. Since we're moving Rails.logger logs to stdout, would this be a good time to do the same for the other ones?

Having had more of a look over things I've now decided not to make this change. I've found the stderr logs are available in Kibana, and I was just failing to find them before. So I don't think I'm really adding any value here (unfortunately). Example:

Screen Shot 2019-07-19 at 08 50 02

(Sidenote: since we regex https://github.com/alphagov/content-publisher/blob/master/app/services/edition_assertions.rb#L29 should we use the regex to just pull out the name of the method?)

I think I got caught out by the message / @message discrepency.

It does still seem like a smell that we use this odd stdout/stderr redirect. It looks like we need to be distinct in what is output as JSON or not, so the best way I could imagine removing the stderr redirect would be if we set logstasher to write to it's own custom file for request logging and then kept stdout for all the normal stuff. 

Looking into this I also spotted an annoyance that we get these long stack traces for when a route isn't matched: 

Screen Shot 2019-07-19 at 08 56 56

PR #144, opened 2020-04-21
Fix spelling of `supressed` => `suppressed`

ChrisBAshton:
This was fixed in logstasher here: https://github.com/shadabahmed/logstasher/commit/bc9e6cfed51a9aea3f27610ee0bcf1d90db5da9b

They've maintained backwards compatibility but it is better if we use the correct parameter, just in
case configuration has shifted. It will also make it easier to look up the correct documentation.

Arises from https://trello.com/c/fX8U5tly/1911-5-fix-whitehall-stacktraces-in-kibana


ChrisBAshton:
Looks like Jenkins can't cope with backticks in branch names (automatically generated from PR title).

PR #149, opened 2020-07-20
Turn on linting

barrucadu:


spec/lib/rails_ext/action_dispatch/debug_exceptions_spec.rb
```diff
@@ -38,11 +40,13 @@ def log_error(request, wrapper); end
       end
       instance = FakeDebugExceptions.new
 
+      # rubocop:disable Lint/AmbiguousBlockAssociation
       expect {
         described_class.monkey_patch_log_error(FakeDebugExceptions)
       }.to change {
         instance.method(:log_error)
```
benthorner, line 47: Minor: I think we can just fix this one, without disabling it (parenthesis for the `to` method).

PR #153, opened 2020-09-04
Add two missing Google domains to Content Security Policy

brucebolt:
We are making requests to www.googletagmanager.com in govuk_publishing_components [1], and including JavaScript and CSS from www.gstatic.com in content-data-admin [2].  All were being blocked by the Content Security Policy, therefore adding these domains to the policy.

1: https://github.com/alphagov/govuk_publishing_components/blob/21e5808e03e7c97721cefe533b315753db7c5403/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb#L16
2: https://github.com/alphagov/content-data-admin/blob/2bbf6cd4bbae29ec02107f4341d526f7707be6d6/app/views/layouts/application.html.erb#L2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+  
+* Add www.googletagmanager.com and www.gstatic.com to Content Security Policy (https://github.com/alphagov/govuk_app_config/pull/153)
+
 # 2.2.1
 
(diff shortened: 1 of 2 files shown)


cbaines (approved):
I don't know much about Google Analytics, but this sounds OK to me

PR #157, opened 2020-09-17
Remove strange whitespace

ChrisBAshton:
Visual Studio Code complained of unusual whitespace and offered
to remove it.


benthorner (approved):
The wrecking ball of an editor strikes again (the previous one, that is) 😆.

PR #159, opened 2020-09-18
Ignore data sync errors and allow custom `should_capture` via Delegator pattern

### LINKED ISSUE #152: Proposal for a generalised approach for data sync error suppression
A big pain point for GOV.UK error monitoring is any code that runs in integration and staging during the nightly database sync. This process drops the current database and then imports a new database, code that executes during this time (such as scheduled tasks) may error if it has to interact with the database. These errors can flood Sentry and make it difficult to spot genuine errors. 

An approach we've taken to working around this scenario in [Content Publisher](https://github.com/alphagov/content-publisher/blob/d088576a1a121d5f1d3492babcf727243e32ab09/config/initializers/govuk_error.rb#L8-L16) and [Email Alert API](https://github.com/alphagov/email-alert-api/blob/f0c61539f7244bdb52db54d277a46412b728d642/config/initializers/govuk_error.rb#L8-L18) is to suppress these errors using the [`should_capture`](https://docs.sentry.io/clients/ruby/config/#optional-settings) callback. Now there is interest in using this for a third application, which seems to indicate a good time to generalise. Naturally it'd be better if we were in a situation where apps didn't run and error during the sync but I'm considering this out of scope for this issue.

Just copying the code from a previous implementation into govuk_app_config seems a flawed approach for a few reasons:

- It only supports Postgres / currently requires postgres
- It clobbers any existing usages of the should_capture callback
- Fetching environment from the sentry environment variable feels a bit tangential
- The data sync times need to be kept in sync with puppet
- It's very hard to test

So to merge this into govuk_app_config I propose that we build a generalised approach.

Ideas for this would be:

## A configuration variable to specify exceptions to exclude during data sync

Similar to how we have [config.excluded_exceptions](https://github.com/alphagov/govuk_app_config/blob/08b3d0659337834313cf757d6f332535dfbb395b/lib/govuk_app_config/configure.rb#L10) it'd be good to have a sibling of `config.govuk_data_sync_excluded_exceptions` where an array of class names can be provided. We could have "PG::Error" included there by default and apps can add their own ones if they like.

To do this we may have to use a [delegator pattern](https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/SimpleDelegator.html) to add extra behaviour to the [Raven config object](https://github.com/alphagov/govuk_app_config/blob/08b3d0659337834313cf757d6f332535dfbb395b/lib/govuk_app_config/govuk_error.rb#L15-L17).

## Create an environment variable for data sync environments

We can replace the hard-coded times and the need to use the `SENTRY_CURRENT_ENV` to determine whether we're in a data sync environment. The contents of this variable can tell us the times it runs from and until and it's mere presence could tell us whether we're in an environment with a data sync period.

As a suggestion we could call it `GOVUK_DATA_SYNC_PERIOD` and have contents such as "22:00-08:00". This could then be parsed and used to work out if we're in a data sync when an error occurs.

Puppet can be used to set this based on [this data](https://github.com/alphagov/govuk-puppet/blob/4bcf837e0dc4f48be59716921841307dfa291ebe/modules/govuk_data_sync_in_progress/manifests/init.pp) hopefully it can be abstracted from the class it is defined in.

## Implementation

With the above the next things we could do is:

- write the code for this in a testable class
- handle parsing the environment variable at app start up time rather than at the point of an error (as a problem with this may make apps break in messy ways if it occurs at runtime)
- work out if we there is a way we can [re-use code from raven](https://github.com/getsentry/raven-ruby/blob/3826f07499d2b1c227d1db394e48c19f1f6e6633/lib/raven/utils/exception_cause_chain.rb) to determine exception cause (the [Content Publisher approach](https://github.com/alphagov/content-publisher/blob/c701fd0c57c47e071f226aa6c817fb45a63506e0/config/initializers/govuk_error.rb#L9-L10) is pretty crude)
- work out how to maintain any existing `should_capture` configuration when initialising

/cc @thomasleese

ChrisBAshton:
In order to prevent thousands of transient errors being reported to Sentry during the data sync window (errors that occur only because databases are copied/provisioned overnight and are sometimes unavailable), we want to configure the `should_capture` logic in govuk_app_config in the same way as [Content Publisher](https://github.com/alphagov/content-publisher/blob/d088576a1a121d5f1d3492babcf727243e32ab09/config/initializers/govuk_error.rb#L8-L16) and [Email Alert API](https://github.com/alphagov/email-alert-api/blob/f0c61539f7244bdb52db54d277a46412b728d642/config/initializers/govuk_error.rb#L8-L18), so that the fix is applied centrally. This involves setting the `should_capture` behaviour to ignore any errors of type `PG::Error`, if we are in the data sync window (10pm to 8am, [according to Puppet](https://github.com/alphagov/govuk-puppet/blob/4bcf837e0dc4f48be59716921841307dfa291ebe/modules/govuk_data_sync_in_progress/manifests/init.pp#L20-L24)).

However, we want to allow downstream apps to be able to define their own `should_capture` conditions too. Therefore, this PR:

* Changes the object apps interact with when they call `GovukError.configure` - it now calls `DefaultConfiguration.configure`, which uses the [Delegator pattern](https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/SimpleDelegator.html) [as suggested by @kevindew](https://github.com/alphagov/govuk_app_config/issues/152). The pattern delegates methods to the `Raven.configuration` object unless they've been explicitly overridden in `DefaultConfiguration`.
* ^ this enables us to define our own default `should_capture=(callback)`, inside `DefaultConfiguration` - but also allows downstream apps to provide their own `should_capture=(callback)` callbacks too. There are loads of ways these variants can be logically combined - described in 61784227ff26c9a7f038c9c64b847ea8e6104599 - but we've gone for a relatively simple `AND`. This means that both the default callback and the custom callback must return `true` for the overall `should_capture` callback to return `true`.
* The default `should_capture` callback examines the error type and determines whether or not we are in the data sync window to decide if the error should be captured. It does this by looking at a global `GOVUK_DATA_SYNC_PERIOD` ENV variable of the form `22:00-08:00`. `PG::Error` errors that occur during this time frame will be ignored.
* The ENV variable doesn't exist yet: I will raise a separate PR in govuk-puppet for that.

This work was quite fiddly. Consequently, there are a couple of tests I haven't figured out how to fix yet, and a slightly meandering commit history. I'm hoping to at least fix the former before the PR is merged.

Trello: https://trello.com/c/rAlvGlSp/2123-5-ignore-data-sync-errors-in-govukappconfig
Fixes #152.


spec/lib/default_configuration_spec.rb
```diff
@@ -1,45 +1,35 @@
...
-      DefaultConfiguration.new(sentry_client)
+      client = DefaultConfiguration.new(Raven.configuration)
+      expect(client.silence_ready).to eq(nil)
     end
 
-    context "we are in a Rails environment" do
-      let!(:cached_rails_env) { Rails.env }
-      after { Rails.env = cached_rails_env }
+    # @TODO - these fail because `defined?(Rails)` returns `nil` for some reason.
+    # (Even if we comment out the `hide_const` code above). Very odd.
```
ChrisBAshton, line 15: 😢

govuk_app_config.gemspec
```diff
@@ -27,11 +27,13 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler", "~> 1.15"
   spec.add_development_dependency "climate_control"
+  spec.add_development_dependency "pry"
```
kevindew, line 30: Any reason for this over byebug? byebug feels more common in our apps.

govuk_app_config.gemspec
```diff
@@ -27,11 +27,13 @@ Gem::Specification.new do |spec|
...
   spec.add_development_dependency "rack-test", "~> 1.1.0"
   spec.add_development_dependency "rails", "~> 6"
   spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "rspec", "~> 3.9.0"
   spec.add_development_dependency "rspec-its", "~> 1.3.0"
   spec.add_development_dependency "rubocop-govuk"
+  spec.add_development_dependency "timecop", "~> 0.9.1"
```
kevindew, line 37: Since we have Rails as a dependency we can use [`ActiveSupport::Testing::TimeHelpers`](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#testing-utilities) to avoid needing an additional dependency.

lib/govuk_app_config/default_configuration.rb
```diff
@@ -0,0 +1,86 @@
+require "govuk_app_config/govuk_data_sync"
+
+class DefaultConfiguration < SimpleDelegator
```
kevindew, line 3: This class should be under the GovukError namespace since there are few things going on in this gem.

I feel like this class has a little too much going on in it with it handling both the default configuration and our logic changes. I'd suggest that you instead have this class as entirely the delegator - So potentially GovukError::Configuration and then have a different file to do the configuration (could re-instate govuk_app_config/configure.rb though I feel like that isn't specific enough that it's govuk_error configuration, so could have a govuk_app_config/govuk_error/configure.rb)

I think that idea would also help in your testing as it seems a pain to test every part of the config (and somewhat superfluous) but you do want to test the logic you're adding. You might even find the data sync code can fit into a shorter version of this class.

lib/govuk_app_config/default_configuration.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+private
+
+  def default_should_capture
+    lambda do |error_or_event|
+      data_sync_ignored_error = error_or_event.is_a?(PG::Error) ||
+        (error_or_event.respond_to?(:cause) && error_or_event.cause.is_a?(PG::Error))
+
+      if !@data_sync.in_progress?
```
kevindew, line 79: Our house style would expect this to be accessed via an attr_reader, it's a shame the linter doesn't nag on it.

lib/govuk_app_config/default_configuration.rb
```diff
@@ -0,0 +1,86 @@
...
+      super(combined)
+    end
+  end
+
+private
+
+  def default_should_capture
+    lambda do |error_or_event|
+      data_sync_ignored_error = error_or_event.is_a?(PG::Error) ||
+        (error_or_event.respond_to?(:cause) && error_or_event.cause.is_a?(PG::Error))
```
kevindew, line 77: Rather than using the `PG::Error` constant directly I think we should have this set-up as a configuration of strings on the config object, similar to `excluded_exceptions` - as otherwise all apps using would need the `PG` constant defined

So i'd imagine you'd have:

```
config.data_sync_excluded_exceptions = ["PG::Error"]
```

Then apps can add their own exceptions too that list from their configuration.

lib/govuk_app_config/default_configuration.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+      super(combined)
+    end
+  end
+
+private
+
+  def default_should_capture
+    lambda do |error_or_event|
+      data_sync_ignored_error = error_or_event.is_a?(PG::Error) ||
```
kevindew, line 76: The Content Publisher implementation of the error detection is pretty crude (only 1 level of cause detection) and I wonder if we can do better - Raven has [this code](https://github.com/getsentry/sentry-ruby/blob/3826f07499d2b1c227d1db394e48c19f1f6e6633/lib/raven/utils/exception_cause_chain.rb) that I think we might be able to re-use.

lib/govuk_app_config/default_configuration.rb
```diff
@@ -0,0 +1,86 @@
+require "govuk_app_config/govuk_data_sync"
+
+class DefaultConfiguration < SimpleDelegator
+  def initialize(raven_configuration_instance)
+    super(raven_configuration_instance)
```
kevindew, line 5: The argument here is redundant by the way, `super` is automatically passed a parent methods arguments.

lib/govuk_app_config/govuk_data_sync.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def in_time_range?(from, to)
+    hour_is_in_range = Time.now.hour >= from.hour || Time.now.hour <= to.hour
+    minute_is_in_range = if Time.now.hour == from.hour
+                           Time.now.min >= from.min
+                         elsif Time.now.hour == to.hour
+                           Time.now.min <= to.min
+                         else
+                           true
+                         end
```
kevindew, line 28: A thing to note is that this will be set to whatever timezone the software running is in, which would be UTC for us. Is that consistent with puppet? (I assume it is)

It might be worth adding a comment to note that the times are expected to be UTC?

lib/govuk_app_config/govuk_data_sync.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class GovukDataSync
+  def initialize(govuk_data_sync_period)
+    @from, @to = false
+    begin
+      parts = govuk_data_sync_period&.split("-")
+      @from, @to = parts.map { |time| Time.parse(time) } if parts&.count == 2
+    rescue ArgumentError
+      # At least one of the parts was malformed. Leave @from/@to as false.
+    end
```
kevindew, line 11: I feel like we might want this to blow up on a malformed env var rather than gracefully handle it. I'm imagining that if we messed up the env var we'd have very little clues about it aside from overnight when the errors started coming in. Whereas if it blew up quickly you'd get much faster feedback.

lib/govuk_app_config/govuk_error.rb
```diff
@@ -11,9 +12,11 @@ def self.notify(exception_or_message, args = {})
     Raven.capture_exception(exception_or_message, args)
   end
 
+  def self.init(raven_configuration = DefaultConfiguration.new(Raven.configuration))
```
kevindew, line 15: I don't think this needs an argument - as I don't think we'd expect anyone to alter it. It'd feel more idiomatic defined within the method, as otherwise it's a slightly confused setter.

lib/govuk_app_config/govuk_error.rb
```diff
@@ -11,9 +12,11 @@ def self.notify(exception_or_message, args = {})
...
+    @raven_configuration = raven_configuration
+  end
+
   def self.configure
-    Raven.configure do |config|
-      yield(config)
-    end
+    yield @raven_configuration
```
kevindew, line 20: I did wonder if we needed the init method at all and could do:
```
Raven.configure do |raven_config|
  yield Configuration.new(raven_config)
end
```

lib/govuk_app_config/govuk_data_sync.rb
```diff
@@ -0,0 +1,31 @@
+require "time"
+
+class GovukDataSync
+  def initialize(govuk_data_sync_period)
+    @from, @to = false
```
kevindew, line 5: nil feels a more idiomatic value than false here. I actually think this is redundant though as I think the begin will have hoisted the ivars already into scope.

P.S. I already wrote a comment on this but it seems to have gone somewhere, apologies if you get two.

lib/govuk_app_config/govuk_data_sync.rb
```diff
@@ -0,0 +1,31 @@
...
+    begin
+      parts = govuk_data_sync_period&.split("-")
+      @from, @to = parts.map { |time| Time.parse(time) } if parts&.count == 2
+    rescue ArgumentError
+      # At least one of the parts was malformed. Leave @from/@to as false.
+    end
+  end
+
+  def in_progress?
+    @from && @to && in_time_range?(@from, @to)
```
kevindew, line 15: Similar to previous comment about attr_reader being the more common idiom in GOV.UK ruby

lib/govuk_app_config/govuk_error.rb
kevindew: On second thoughts, it is better with an `.init` method since that'll preserve state in our delegator even if there are multiple `.configure` calls. So keep doing what you're doing here.

kevindew (commented):
This is very exciting Chris, nice job - I'm looking forward to it. I've added a bunch of comments on the implementation side, I hope their helpful - there are a few implementation nuances to ponder.

I didn't really dig into the tests or commits at this stage.

lib/govuk_app_config/govuk_error.rb
ChrisBAshton: Thanks - I initially intended for it to be easily swapped in for testing purposes, in an Open for Extension/Closed for Modification principle. But as I went on, the tests changed enough that it wasn't strictly needed anymore.

I seem to remember attempting to move it into the method itself, but that caused one failing test (in `configuring_spec.rb`). I'll have another go though 👍

lib/govuk_app_config/govuk_error.rb
```diff
@@ -1,4 +1,5 @@
 require "sentry-raven"
+require "govuk_app_config/govuk_error/configuration"
```
ChrisBAshton, line 2: Note to self: recode/rebase so that the delegator pattern comes _first_

_then_ the code to add hardcoded data sync window inside the should_capture

_then_ the code to generalise that in an ENV.

ChrisBAshton:
Superseded by #160.

ChrisBAshton:
Thanks for the review, @kevindew - it gave me a lot of food for thought!

I've addressed your comments and raised a new PR at #160.

PR #160, opened 2020-10-01
Ignore data sync errors and allow custom `should_capture` via Delegator pattern

### LINKED ISSUE #152: Proposal for a generalised approach for data sync error suppression

ChrisBAshton:
In order to prevent thousands of transient errors being reported to Sentry during the data sync window (errors that occur only because databases are copied/provisioned overnight and are sometimes unavailable), we want to configure the `should_capture` logic in govuk_app_config in the same way as [Content Publisher](https://github.com/alphagov/content-publisher/blob/d088576a1a121d5f1d3492babcf727243e32ab09/config/initializers/govuk_error.rb#L8-L16) and [Email Alert API](https://github.com/alphagov/email-alert-api/blob/f0c61539f7244bdb52db54d277a46412b728d642/config/initializers/govuk_error.rb#L8-L18), so that the fix is applied centrally. This involves setting the `should_capture` behaviour to ignore any errors of type `PG::Error`, if we are in the data sync window (10pm to 8am, [according to Puppet](https://github.com/alphagov/govuk-puppet/blob/4bcf837e0dc4f48be59716921841307dfa291ebe/modules/govuk_data_sync_in_progress/manifests/init.pp#L20-L24)).

However, we want to allow downstream apps to be able to define their own `should_capture` conditions too. Therefore, this PR:

* Changes the object apps interact with when they call `GovukError.configure` - it now yields to a `RavenDelegator`, which uses the [Delegator pattern](https://ruby-doc.org/stdlib-2.5.1/libdoc/delegate/rdoc/SimpleDelegator.html) [as suggested by @kevindew](https://github.com/alphagov/govuk_app_config/issues/152). The pattern delegates methods to the `Raven.configuration` object unless they've been explicitly overridden.
* ^ this enables us to define our own default `should_capture=(callback)`, inside `DefaultConfiguration` - but also allows downstream apps to provide their own `should_capture=(callback)` callbacks too. In our implementation, both the default callback and the custom callback must return `true` for the overall `should_capture` callback to return `true`.
* The default `should_capture` callback examines the error type and determines whether or not we are in the data sync window to decide if the error should be captured. It does this by looking at a global `GOVUK_DATA_SYNC_PERIOD` ENV variable of the form `22:00-08:00`. `PG::Error` errors that occur during this time frame will be ignored.
* The ENV variable doesn't exist yet: I will raise a separate PR in govuk-puppet for that.

Trello: https://trello.com/c/rAlvGlSp/2123-5-ignore-data-sync-errors-in-govukappconfig
Fixes #152. Supersedes #159 (the code has been largely recreated by hand here, as the commit history was meandering and no longer particularly valuable).


govuk_app_config.gemspec
```diff
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency "statsd-ruby", "~> 1.4.0"
   spec.add_dependency "unicorn", ">= 5.4", "< 5.8"
 
-  spec.add_development_dependency "bundler", "~> 1.15"
+  spec.add_development_dependency "activesupport"
```
kevindew, line 28: This isn't needed because it's part of Rails which is already a dependency.

govuk_app_config.gemspec
```diff
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
...
+  spec.add_development_dependency "bundler"
```
kevindew, line 29: I wonder if we benefit from defining this as a dependency at all, our other gems don't seem to have it and you need bundler before you can install gems. I'd see if there are any consequences to removing it.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
```diff
@@ -0,0 +1,56 @@
...
+    class MissingDataSyncPeriod < StandardError
+      def message
+        "You must specify a ENV['GOVUK_DATA_SYNC_PERIOD'] (should be of form '22:00-03:00')."
+      end
+    end
+
+    attr_reader :from, :to
+
+    def initialize(govuk_data_sync_period)
+      raise MissingDataSyncPeriod if govuk_data_sync_period.nil?
```
kevindew, line 26: I don't think we should raise an error on this and should use whether the env var is defined or not to determine whether we're doing this.

My logic is that all our apps will have the same code running in prod, staging and integration and the differentiator between prod and the others will be the presence of this env var.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
```diff
@@ -0,0 +1,56 @@
...
+      raise MalformedDataSyncPeriod, govuk_data_sync_period
+    end
+
+    def in_progress?
+      from.present? && to.present? && in_time_range?(from, to)
+    end
+
+  private
+
+    # `from`/`to` times are expected in UTC, based on:
```
kevindew, line 42: Thanks for adding this :-) might be good to tweak it slightly to "`from`/`to` times are done in the servers local time which is expected to be UTC, based on:"

lib/govuk_app_config/govuk_error/raven_delegator.rb
```diff
@@ -0,0 +1,17 @@
+module GovukError
+  class RavenDelegator < SimpleDelegator
+    def should_capture=(closure)
+      # rubocop:disable Style/RedundantSelf
```
kevindew, line 4: Oh how come you needed this?

lib/govuk_app_config/govuk_error/configure_defaults.rb
```diff
@@ -0,0 +1,82 @@
...
+      @data_sync_excluded_exceptions = ["PG::Error"]
+      config.should_capture = lambda do |error_or_event|
+        exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+        data_sync_ignored_error = false
+        data_sync_excluded_exceptions.each do |exception_to_ignore|
+          break if data_sync_ignored_error
+
+          data_sync_ignored_error = exception_chain.reduce(false) do |memo, exception|
+            memo || exception.class == exception_to_ignore ||
+              (exception.respond_to?(:cause) && exception.cause.class == exception_to_ignore)
```
kevindew, line 70: I don't think we need to be looking at the cause here too as that should be covered by the exception chain.

lib/govuk_app_config/govuk_error/configure_defaults.rb
```diff
@@ -0,0 +1,82 @@
...
+      data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
+      @data_sync_excluded_exceptions = ["PG::Error"]
+      config.should_capture = lambda do |error_or_event|
+        exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+        data_sync_ignored_error = false
+        data_sync_excluded_exceptions.each do |exception_to_ignore|
+          break if data_sync_ignored_error
+
+          data_sync_ignored_error = exception_chain.reduce(false) do |memo, exception|
+            memo || exception.class == exception_to_ignore ||
```
kevindew, line 69: I think we've got an issue here where exception.class will be a class object whereas exception_to_ignore will be a string?

lib/govuk_app_config/govuk_error/configure_defaults.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
+      @data_sync_excluded_exceptions = ["PG::Error"]
+      config.should_capture = lambda do |error_or_event|
+        exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+        data_sync_ignored_error = false
+        data_sync_excluded_exceptions.each do |exception_to_ignore|
+          break if data_sync_ignored_error
+
+          data_sync_ignored_error = exception_chain.reduce(false) do |memo, exception|
```
kevindew, line 68: Minor, you might find an any? or none? method serves better here.

kevindew (commented):
Thanks Chris, lots of great improvements here and this is shaping up great. There's a few areas where I think the suggestions I've made didn't come out too clearly so I did some hacking myself so I could try point to what I meant rather than just try describe it in words - hopefully that provides greater clarity.

I cobbled my bits together here: https://github.com/alphagov/govuk_app_config/commit/81c78c576d55f3f6c261691ca6b9d90a2abbfc66

So these are the bits that perhaps I think our wires ended up a bit crossed on.

I think we should make the data sync error side distinct from the configuration. I've tried applying that like so: https://github.com/alphagov/govuk_app_config/commit/81c78c576d55f3f6c261691ca6b9d90a2abbfc66#diff-e455739649a523940d6e5fc5353656a0R3-R36 - that also reflects a difference on should capture, I'd expect an application defining should_capture will replace a previously defined version, but always keep the data sync check.

https://github.com/alphagov/govuk_app_config/commit/81c78c576d55f3f6c261691ca6b9d90a2abbfc66#diff-60bf2f76f5886144ddb9f524d7a5a8edR49-R51 represents how I'd expect that errors were configured, so they rather transparently fit into configuration rather than containing some of the logic. I also felt it didn't feel quite right to be setting our default configuration in and amongst a class, seems to raise questions about what to test or not.

I did have a quick pass at matching to the exception class by converting the names to strings here: https://github.com/alphagov/govuk_app_config/commit/81c78c576d55f3f6c261691ca6b9d90a2abbfc66#diff-e455739649a523940d6e5fc5353656a0R30-R32 to be honest I don't know if that is sufficient or not as Raven seems to have used a more [complex constant resolving process](https://github.com/getsentry/sentry-ruby/blob/7a0cebc54c5508d379769c7af688c8218721d734/lib/raven/configuration.rb#L446-L448). It seemed to work in my test though. I guess a thing `to_s` might not cover is two constants pointing to the same object, but that seemed pretty edge case. 

Hope this helps and happy to have a chat if some of my suggestions are ambiguous, there are a variety of subtle things in my hacking that I haven't explained.

lib/govuk_app_config/govuk_error/raven_delegator.rb
ChrisBAshton: Ah, I needed it in an earlier iteration of the code, but have just removed it and it works fine :) good spot.

lib/govuk_app_config/govuk_error/configure_defaults.rb
ChrisBAshton: In my tests it would still pass whether or not a `.to_s` is appended to `.class`. I think the class equality method does some automatic conversion to a string if that's what it's being compared to. Perhaps a `.to_s` would reduce confusion, even if it's slightly verbose. I'll add it in.

lib/govuk_app_config/govuk_error/configure_defaults.rb
ChrisBAshton: That's _much_ nicer. Thanks!

lib/govuk_app_config/govuk_error/configure_defaults.rb
ChrisBAshton: D'oh! Good point. Removed!

ChrisBAshton:
Thanks again for the review, @kevindew. Apologies it's taken a while to address your comments - other things have taken priority this week.

I've administered your suggestions in the relevant commits, and then added a final commit for moving the data sync logic into the RavenDelegator.

I'm aware you've opted for a `configure.rb` for the default configuration and a `configuration.rb` for the should_capture logic. I've left as-is for now (`configure_defaults.rb` and `raven_delegator.rb` respectively) as I think this is more scannable, but I'm happy to be talked out of it. (I'm personally of the view that 'configuration' implies where the default config lives, and I wouldn't know what 'configure' is supposed to do in that context).

spec/govuk_error/raven_delegator_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      travel_to(Time.current.change(hour: 23)) do
+        expect(client.should_capture.call(StandardError.new)).to eq(false)
+      end
+    end
+  end
+
+  describe ".should_capture=" do
+    it "Allows apps to add their own `should_capture` callback, that is evaluated alongside the default. If both return `true`, then we should capture, but if either returns `false`, then we shouldn't." do
+      class CustomErrorToIgnore < StandardError; end
```
kevindew, line 68: I think this can have surprising scoping behaviour applied in RSpec (i think it comes out with global scope). The better approach is to use `Class.new(StandardError)` which could be stubbed as a constant for the test with `stub_const("CustomErrorToIgnore", Class.new(StandardError))` : https://relishapp.com/rspec/rspec-mocks/docs/mutating-constants

To be honest though I don't think you need to actually go to the level of stubbing constants as should_capture doesn't necessarily take an exception so you could do:

```
raven_configurator.should_capture = lambda do |error_or_event|
  error_or_event == "do capture"
end

expect(raven_configurator.should_capture.call("do capture")).to be_true
expect(raven_configurator.should_capture.call("don't capture")).to be_false
```

lib/govuk_app_config/govuk_error/configure_defaults.rb
```diff
@@ -0,0 +1,68 @@
+module GovukError
+  class ConfigureDefaults < SimpleDelegator
```
kevindew, line 2: This doesn't need to be a simple delegator or have attributes as it is just calling methods on the config instance

kevindew (commented):
Thanks Chris, this is looking really close now. Nice job.

Regarding the areas you referenced in your previous comment.

ConfigureDefaults class, to me this isn't a class in a traditional sense all we're looking for is vessel for the configuration code, and it doesn't really behave like a class since we just use it as a function to apply side effects to the argument.

I put it in a govuk_error/configure file to reference that's it's just to configure that module. It could well go inside the block of https://github.com/alphagov/govuk_app_config/pull/160/files#diff-de4bdf6f6685d2a23ac3a9ea5c54b985R10-R12 instead - yeah my concerns are mostly about this being a class.

With the naming switch from GovukError::RavenDelegator to GovukError::Configuration, my logic for that was that the object is a configuration object for GovukError, whereas it does more than be a delegator to Raven. That naming thought more like it was describing the implementation than the purpose.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
```diff
@@ -0,0 +1,50 @@
...
+      hour_is_in_range = Time.now.hour >= from.hour || Time.now.hour <= to.hour
+      minute_is_in_range = if Time.now.hour == from.hour
+                             Time.now.min >= from.min
+                           elsif Time.now.hour == to.hour
+                             Time.now.min <= to.min
+                           else
+                             true
+                           end
+      hour_is_in_range && minute_is_in_range
+    end
```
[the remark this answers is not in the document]
ChrisBAshton: That sounds ace! However, I've tried it and it fails for this test (for the range `"22:30-8:30"`):

```
  1) GovukError::GovukDataSync.in_progress? returns true if we are within the time range
     Failure/Error: at(hour: 22, min: 31) { expect(data_sync.in_progress?).to eq(true) }
     
       expected: true
            got: false
     
       (compared using ==)
     
       Diff:
       @@ -1,2 +1,2 @@
       -true
       +false
```

(also fails if I change it to 22:31. I'm guessing it doesn't account for when the `from` is a later time than the `to`).

spec/govuk_error/raven_delegator_spec.rb
ChrisBAshton: Thanks for the tip 👍  much nicer!

ChrisBAshton:
Thanks @kevindew - some excellent suggestions there, have applied them all.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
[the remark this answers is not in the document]
kevindew: Oh yes sorry I forgot that they're both relative to current date 🤦  - it's annoying there isn't a time class that is decoupled from dates.

spec/govuk_error/configuration_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+      end
+    end
+
+    it "should ignore exceptions whose underlying cause is an ignorable error, if it occurred during the data sync" do
+      pg_error = double("Caused by PG::Error", class: "PG::Error")
+      allow(pg_error).to receive(:cause)
+      exception = double("Exception 1", cause: double("Exception 2", cause: pg_error))
+
+      configuration.data_sync_excluded_exceptions << pg_error.class
```
kevindew, line 42: Should `pg_error.class` be `"PG::Error"` ? I notice it's not a string (which, incidentally, if this is allowed behaviour may want to be a separate test)

spec/integration/configuring_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe "Requiring govuk_app_config" do
   it "correctly initialises error tracking" do
-    ClimateControl.modify SENTRY_CURRENT_ENV: "integration-or-somesuch" do
+    ClimateControl.modify SENTRY_CURRENT_ENV: "integration-or-somesuch", GOVUK_DATA_SYNC_PERIOD: "22:00-08:00" do
```
kevindew, line 5: Is this change necessary?

lib/govuk_app_config/govuk_error.rb
```diff
@@ -12,8 +13,6 @@ def self.notify(exception_or_message, args = {})
   end
 
   def self.configure
-    Raven.configure do |config|
-      yield(config)
-    end
+    yield Configuration.new(Raven.configuration)
```
kevindew, line 16: I think configuration would benefit from being saved as a variable - otherwise if you run configure twice it'll keep all the Raven configuration but lose any of our custom configuration.

spec/lib/govuk_error_spec.rb
```diff
@@ -19,4 +19,15 @@
...
     end
   end
+
+  describe ".configure" do
+    it "configures Raven via the Configuration" do
+      mock_configuration = double("Configuration")
+      allow(GovukError::Configuration).to receive(:new) { mock_configuration }
+      expect(mock_configuration).to receive(:foo)
+      # rubocop:disable Style/SymbolProc
+      GovukError.configure { |config| config.foo }
```
kevindew, line 29: It doesn't seem too necessary to turn rubocop off here. Could just be `GovukError.configure(&:foo)`. If you want something more representative of the class you could stub a `foo=` method and then have a `config.foo = "something"` which won't need the rubocop disabling.

lib/govuk_app_config/govuk_error/configure.rb
```diff
@@ -1,11 +1,13 @@
...
+  config.current_environment = ENV["SENTRY_CURRENT_ENV"]
+
   config.before_send = proc { |e|
     GovukStatsd.increment("errors_occurred")
     GovukStatsd.increment("error_types.#{e.class.name.demodulize.underscore}")
     e
   }
 
-  config.silence_ready = !Rails.env.production? if defined?(Rails)
+  config.silence_ready = !Rails.env.production? if Object.const_defined?("Rails")
```
kevindew, line 10: Do we need to make this change? I recall it being from an earlier fight with tests that seems to have gone.

lib/govuk_app_config/govuk_error/configure.rb
```diff
@@ -1,11 +1,13 @@
 GovukError.configure do |config|
+  config.current_environment = ENV["SENTRY_CURRENT_ENV"]
```
kevindew, line 2: Is this needed?

kevindew (commented):
Great thanks Chris - I think we're through all the meat of this now, just a few minor things. 

Have you given it a try out on an app? You could try install with Content Publisher I did that myself for a quick spin earlier in the process.

spec/integration/configuring_spec.rb
ChrisBAshton: Looks like it's not needed anymore (it was at one point). Thanks!

lib/govuk_app_config/govuk_error/configure.rb
ChrisBAshton: Without it, I get a failing test:

```
Failures:

  1) Requiring govuk_app_config correctly initialises error tracking
     Failure/Error: expect(Raven.configuration.current_environment).to eql("integration-or-somesuch")
     
       expected: "integration-or-somesuch"
            got: "default"
     
       (compared using eql?)
     # ./spec/integration/configuring_spec.rb:8:in `block (3 levels) in <top (required)>'
     # ./spec/integration/configuring_spec.rb:5:in `block (2 levels) in <top (required)>'
```

spec/govuk_error/configuration_spec.rb
ChrisBAshton: Good spot - I was under the impression `.class` returned the name of the class as a string (and of course does in this case, as it's a double). It was a way of DRYing up the test a little but now I see it was misleading, as trying `.class` in IRB just returns `Class`...
Have changed to "PG::Error" 👍

lib/govuk_app_config/govuk_error.rb
ChrisBAshton: I could change it to this:

```ruby
  def self.configure
    @configuration ||= Configuration.new(Raven.configuration)
    yield @configuration
  end
```

However, it is easier said than done - I have the same problematic test [as below](https://github.com/alphagov/govuk_app_config/pull/160#discussion_r504475270). By the time this test is run, the `environment` has already been set and is no longer looking at the `SENTRY_CURRENT_ENV` ENV variable. I'll need to have a think about how to get this test working.

lib/govuk_app_config/govuk_error/configure.rb
kevindew: Hmm the curious thing to me is that we weren't setting current environment before and I assume this already worked? Is the failing test due to Raven configuration being a singleton and maintaining state?

lib/govuk_app_config/govuk_error.rb
kevindew: Oh sorry should have read this comment before commenting on the last one - oh well. It now makes sense to me.

So behaviour wise setting the instance variable seems the right thing to do but yes all class instance variables are a real pain with testing.

Given we have the same singleton behaviour with raven I think it might be better to rethink the testing strategy to accommodate it. So we could test that it yields a configuration instance? After a [quick google](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/yield-matchers) perhaps it could be `expect { |b| GovukError.configure(&b) }.to yield_once_with_args(instance_of(GovukError::Configuration))`

ChrisBAshton: Thanks Kevin - I hadn't found time to look at this earlier, but have now implemented the test change. I've encapsulated it in its own commit at the end of the PR.

README.md
```diff
@@ -89,6 +89,27 @@ GovukError.configure do |config|
 end
 ```
 
+And you can exclude errors from being reported if they occur during the nightly data sync:
```
kevindew, line 92: Might want to add a quick qualifier that this applies to integration and staging.

lib/govuk_app_config/govuk_error/configure.rb
```diff
@@ -5,7 +5,7 @@
     e
   }
 
-  config.silence_ready = !Rails.env.production? if defined?(Rails)
+  config.silence_ready = !Rails.env.production? if defined?("Rails")
```
kevindew, line 8: Did you mean to make Rails a string here?

spec/govuk_error/govuk_data_sync_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+          "\"#{val}\" is not a valid value (should be of form '22:00-03:00').",
+        )
+      end
+    end
+  end
+
+  describe ".in_progress?" do
+    it "returns false if we are outside of the time range" do
+      data_sync = GovukError::GovukDataSync.new("22:30-8:30")
+      at(hour: 21) { expect(data_sync.in_progress?).to eq(false) }
```
kevindew, line 37: This is nice 👍

kevindew (approved):
Awesome work Chris 💯  🏅 . I've got two very minor comments but this shouldn't need an extra review.

lib/govuk_app_config/govuk_error/configure.rb
ChrisBAshton: Oops - good spot!

kevindew (approved):

ChrisBAshton:
(@kevindew I've just pushed a 'bump to 2.5.0' commit, which involved a rebase as the version had changed while this PR was open. There are no other new changes).

PR #163, opened 2020-10-20
WIP: Provide custom fingerprinting algorithm for Sentry

ChrisBAshton:
Process we intend to code up:

1. Reduce stack trace to "application traces" only
2. If no stack trace exists, return the error name instead
3. With support for exception chains. (We then need to think
   about whether it makes sense to reorder them such that the
   originating exception is reported first).
4. Think about where transactions fit into this - they look
   like they were taken into account in the audit.
   https://github.com/getsentry/sentry-ruby/blob/fbbc7a51ed10684d0e8b7bd9d2a1b65a7351c9ef/sentry-ruby/lib/sentry/event.rb#L24

Justification for 1):
- Stack traces that contain "framework traces" are highly
  sensitive to changes in Ruby version, dependencies etc,  and
  thus we often have multiple errors reported in Sentry that
  should really be grouped as one.
- However, we don't want to use Rails::BacktraceCleaner in the
  `backtrace_cleanup_callback` because it's lossy - it removes
  the framework traces altogether, prior to it even getting into
  Sentry, so we lose valuable diagnostic information. We only
  want to remove framework traces for the purposes of
  fingerprinting (grouping); we should still be able to dig into
  the original stack trace when we need to.

Trello: https://trello.com/c/NZNjFHWO/2162-5-improve-sentrys-grouping-of-exceptions

### DIFF
--- a/lib/govuk_app_config/govuk_error/configure.rb
+++ b/lib/govuk_app_config/govuk_error/configure.rb
@@ -7,6 +7,11 @@ GovukError.configure do |config|
 
   config.silence_ready = !Rails.env.production? if defined?(Rails)
 
+  config.before_send = lambda { |event, hint|
+    event.fingerprint = GovukError::CustomFingerprint.sample(event)
+    event
+  }
+
   config.excluded_exceptions = [
     # Default ActionDispatch rescue responses
     "ActionController::RoutingError",
--- a/lib/govuk_app_config/govuk_error/custom_fingerprint.rb
+++ b/lib/govuk_app_config/govuk_error/custom_fingerprint.rb
@@ -1,0 +1,10 @@
+module GovukError
+  class CustomFingerprint
+    def self.sample(event)
+      # TODO - it's not clear if this is possible, given that Ruby SDK fingerprinting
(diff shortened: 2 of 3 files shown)


lib/govuk_app_config/configure.rb
```diff
@@ -7,6 +7,11 @@
 
   config.silence_ready = !Rails.env.production? if defined?(Rails)
 
+  config.before_send = lambda { |event, hint|
+    event.fingerprint = ['fingerprint'] # STACK TRACE
```
ChrisBAshton, line 11: Example from email-alert-api:

```
{"exception_class":"ActiveRecord::RecordNotFound","exception_message":"Couldn't find Subscriber","stacktrace":["app/models/subscriber.rb:26:in `find_by_address!'","app/controllers/subscribers_auth_token_controller.rb:18:in `find_subscriber'","app/controllers/subscribers_auth_token_controller.rb:5:in `auth_token'"]}
```

ChrisBAshton:
Making a change of tack. Rewriting the fingerprinting algorithm feels risky and will inevitably involve quite a lot of trial and error. I've opened an issue on sentry-raven instead, proposing a cleaner approach:

https://github.com/getsentry/sentry-ruby/issues/1076

PR #165, opened 2020-10-21
Also ignore subclasses of `data_sync_excluded_exceptions`

ChrisBAshton:
We're seeing PG:UndefinedTable errors in Sentry, which we'd like
to ignore during the data sync. By default we're ignoring PG::Error
exceptions already, but this is a specific sub class of PG::Error
(and there are many more we envisage wanting to ignore). We
could list all of these subclasses in the
`data_sync_excluded_exceptions` array, but what we'd rather do
is describe the parent class and have any of its children
exceptions ignored.

`PG::Error === PG::UndefinedTable.new` is true, whereas
`"PG::Error" === PG::UndefinedTable.new.class` is false.
This commit makes the default `should_capture` behaviour dig
deeper into the exception, checking that the exception type
is classed as any one of the errors we've chosen to ignore.

From now on, any exceptions that are a subclass of `PG::Error`
will be ignored during the data sync.

https://trello.com/c/rAlvGlSp/2123-5-ignore-data-sync-errors-in-govukappconfig

### DIFF
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -25,8 +25,13 @@ module GovukError
     def ignore_excluded_exceptions_in_data_sync
       lambda { |error_or_event|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
+          exception_to_ignore = Object.const_get(exception_to_ignore)
           exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
-          exception_chain.any? { |exception| exception.class.to_s == exception_to_ignore }
+          exception_chain.any? { |exception| exception.is_a?(exception_to_ignore) }
+        rescue NameError
+          # the exception type represented by the exception_to_ignore string
+          # doesn't even exist in this environment, so won't be found in the chain
+          false
         end
 
         !(data_sync.in_progress? && data_sync_ignored_error)
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -25,7 +25,7 @@ module GovukError
     def ignore_excluded_exceptions_in_data_sync
       lambda { |error_or_event|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
-          exception_to_ignore = Object.const_get(exception_to_ignore)
+          exception_to_ignore = Object.const_get(exception_to_ignore) unless exception_to_ignore.is_a?(Module)
           exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
           exception_chain.any? { |exception| exception.is_a?(exception_to_ignore) }
         rescue NameError
--- a/spec/govuk_error/configuration_spec.rb
+++ b/spec/govuk_error/configuration_spec.rb
@@ -12,6 +12,8 @@ RSpec.describe GovukError::Configuration do
   end
 
   describe ".should_capture" do
+    let(:configuration) { GovukError::Configuration.new(Raven.configuration) }
+
     context "during the data sync" do
       around do |example|
         ClimateControl.modify GOVUK_DATA_SYNC_PERIOD: "22:00-08:00" do
@@ -20,8 +22,6 @@ RSpec.describe GovukError::Configuration do
           end
         end
       end
-
-      let(:configuration) { GovukError::Configuration.new(Raven.configuration) }
 
       it "should capture errors by default" do
         expect(configuration.should_capture.call(StandardError.new)).to eq(true)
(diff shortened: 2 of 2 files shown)


lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -25,8 +25,13 @@ def should_capture=(closure)
     def ignore_excluded_exceptions_in_data_sync
       lambda { |error_or_event|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
+          exception_to_ignore = Object.const_get(exception_to_ignore)
```
kevindew, line 28: For compatibility with the Raven interface this should be able to handle the actual class passed in. https://github.com/getsentry/sentry-ruby/blob/7a0cebc54c5508d379769c7af688c8218721d734/lib/raven/configuration.rb#L447

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -25,8 +25,13 @@ def should_capture=(closure)
...
           exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
-          exception_chain.any? { |exception| exception.class.to_s == exception_to_ignore }
+          exception_chain.any? { |exception| exception.is_a?(exception_to_ignore) }
```
kevindew, line 30: I notice raven relies on the `===` method - though `is_a?` is a lot more readable and I can't imagine any reliance on subtle nuances between them. Thought I'd note it though since we got a little burnt last time I thought there wasn't a big consequence diverging from Raven 🤕

spec/govuk_error/configuration_spec.rb
```diff
@@ -12,36 +12,68 @@
...
-      configuration.data_sync_excluded_exceptions << "StandardError"
+      it "should ignore errors that have been added to data_sync_excluded_exceptions" do
+        configuration.data_sync_excluded_exceptions << "StandardError"
 
-      travel_to(Time.current.change(hour: 23)) do
         expect(configuration.should_capture.call(StandardError.new)).to eq(false)
       end
+
+      it "should ignore errors whose underlying cause is an exception in data_sync_excluded_exceptions" do
+        stub_const("PG::Error", Class.new(StandardError))
```
kevindew, line 37: This `PG::Error` stands out to me as odd as all the other classes we're using for exceptions are made up ones. I'd suggest it being another made up one for consistency with the rest of the test.

spec/govuk_error/configuration_spec.rb
```diff
@@ -12,36 +12,69 @@
...
+          begin
+            raise PG::Error
+          rescue PG::Error
+            raise SomeOtherError
+          end
+        rescue SomeOtherError => e
+          chained_exception = e
+        end
+
+        expect(chained_exception).to_not be_nil
```
kevindew, line 52: I guess this is added because the test would still pass if chained_exception had its initial value. I think if you're going to assert on it you may as well assert that it is the class you're expecting. Eg `expect(chained_exception).to be_an_instance_of(SomeOtherError)`

kevindew (commented):
Looks great - thanks for fixing so quickly

I wonder if we can test this branch out directly in Email Alert API or similar to verify it works for real before doing a release? Just would seem a shame if we hit a different problem and had to re-release gem again.

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: This was quite simple to add, so I've tacked a commit on - but I do wonder how necessary it is, given `data_sync_excluded_exceptions` is our own construct, and the [Raven docs](https://docs.sentry.io/platforms/ruby/configuration/options/) seem to suggest passing string values to `excluded_exceptions` (the closest thing to what we've built).

Happy to remove the commit I've just added, or keep it - I don't feel strongly either way.

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: Good to note - but I think it's worth taking the risk, as it does seem more readable given it handles subclasses etc (`===` seems quite misleading in that regard).

ChrisBAshton:
Thanks for the review, @kevindew. The only way I can think of testing it would be to make a branch of email-alert-api which pulls in this branch instead of the govuk_app_config gem, building that to integration, and waiting for the data sync to happen overnight (hoping nobody builds email-alert-api in the meantime). Is that what you suggest?

Alternatively I guess I could edit `/etc/govuk/env.d/GOVUK_DATA_SYNC_PERIOD` to `00:00-23:59` and then run  `sudo -u govuk-backup /usr/local/bin/govuk_env_sync.sh -f /etc/govuk_env_sync/pull_email_alert_api_staging_daily.cfg` to speed things along...?

lib/govuk_app_config/govuk_error/configuration.rb
kevindew: Cool thanks - I think it's better that the APIs are compatible. It probably isn't absolutely necessary but does seem a common pattern where classes can be passed in directly or as a string, feels unusual to only accept string but not the actual constant.

kevindew (approved):
Great thanks @ChrisBAshton 

With regards to testing we could always merge Email Alert API to be pointing at the git source rather than a Rubygems release for overnight testing - less brittle than a branch deploy.

But I think the env var idea is better, you shouldn't need to do the whole data sync though you just need something that'll generate a `PG::Error` for example: `ActiveRecord::Base.connection.execute("SELECT * FROM blah")`

ChrisBAshton:
Tested on integration with your help (`GovukError.notify(PG::Error.new)`). Prior to this commit, errors went through to Sentry. Now they get silenced:

```
irb(main):001:0> ENV["GOVUK_DATA_SYNC_PERIOD"]
=> "0:0-23:59"
irb(main):002:0> GovukError.notify(PG::UndefinedTable.new)
=> false
irb(main):003:0> GovukError.notify(PG::Error.new)
=> false
```

PR #167, opened 2020-10-22
Fix govuk_app_config in Ruby 2.7 environments by explicitly requiring the 'delegate' library

ChrisBAshton:
[Explanation](https://stackoverflow.com/questions/60459709/where-is-simpledelegator-in-ruby-2-7).
TLDR:

> The Delegator and SimpleDelegator classes aren't core classes
> like Array or Mutex. They're part of the delegate standard
> library which needs to be loaded first: require 'delegate'.
> It happened to work in older Ruby versions as they came with an
> older RubyGems version by default. RubyGems is automatically
> loaded since Ruby 1.9 and until 3.1.0 that meant delegate was
> loaded indirectly. Updating RubyGems or running ruby with
> --disable=gems should cause the exact same issue with Ruby <=
> 2.6 too. irb also loads several standard libraries: delegate
> but also timeout and many more.

This is causing the [bouncer build to error](https://github.com/alphagov/bouncer/pull/273).

### DIFF
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -1,3 +1,4 @@
+require "delegate"
 require "govuk_app_config/govuk_error/govuk_data_sync"
 
 module GovukError
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.5.2
+
+* Fix govuk_app_config in Ruby 2.7 environments by explicitly requiring the 'delegate' library (https://github.com/alphagov/govuk_app_config/pull/167)
+
 # 2.5.1
 
 * Increase scope of `data_sync_excluded_exceptions` so that it includes subclasses (https://github.com/alphagov/govuk_app_config/pull/165)
--- a/lib/govuk_app_config/version.rb
+++ b/lib/govuk_app_config/version.rb
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "2.5.1".freeze
+  VERSION = "2.5.2".freeze
 end


deborahchua (approved):

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -1,3 +1,4 @@
+require "delegate"
```
thomasleese, line 1: Do we need to explicitly [add it as a dependency in the gemspec](https://github.com/alphagov/govuk_app_config/blob/d041c5a868f092d196c126523cc1593a601da89b/govuk_app_config.gemspec#L23-L26) as well? i.e. `spec.add_dependency 'delegate'`

I imagine that in most cases this is coming in already via Rails, but being explicit means that it's guaranteed to always be available.

Although I may have misunderstood something though because https://github.com/ruby/delegate#installation suggests it needs to be in the `Gemfile`, but if it's part of the standard library, I wouldn't expect you to need to do that...

ChrisBAshton: Good point. I hit 'merge' too soon. But I think it's safe because I tested it here:  https://github.com/alphagov/bouncer/pull/273/commits/1f6f705c396e3e126e8f31076a1fa45430c6eaf3

kevindew: It's one of the default gems bundled with Ruby https://stdgems.org/ so is available by default.

I'm not sure what the expected etiquette is for gems that depend on stdgems though, it does seem prudent that they're specified as a dependency as you don't know if a later Ruby will drop them but it also feels unconventional. It's probably worth checking out some of the bigger gems for this next time we ponder about this and see if there is a precedent forming.

PR #171, opened 2020-11-10
Remove Remote IP middleware for all apps

benthorner:
https://trello.com/c/7qWgBlA7/617-investigate-handling-actiondispatchremoteipipspoofattackerror

This is unreliable and we only use it in Signon, although even there
it's unclear why it's necessary [1]. Recently we have seen this lead
to Sentry issues where the middleware irrelevantly detects a dodgy
set of headers in the request [2].

[1]: https://github.com/alphagov/signon/pull/1134
[2]: https://sentry.io/organizations/govuk/issues/1703518396/?project=202221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.8.0
+
+* Turn off redundant RemoteIp middleware for all apps
+
 # 2.7.0
 
 * Ignore intermittent template retrieval errors from Slimmer
--- a/lib/govuk_app_config/railtie.rb
+++ b/lib/govuk_app_config/railtie.rb
@@ -3,5 +3,14 @@ module GovukAppConfig
     config.before_initialize do
       GovukLogging.configure if Rails.env.production?
     end
+
+    # Avoid relying on dodgy HTTP headers about the requester's IP:
+    # https://blog.gingerlime.com/2012/rails-ip-spoofing-vulnerabilities-and-protection/
+    #
+    # Protection of GOV.UK should be implemented at the network level:
+    # https://github.com/alphagov/govuk-cdn-config/blob/2c8b87fd6d1bef7067ea872f7232c53effbf31b4/vcl_templates/www.vcl.erb#L203
+    initializer "govuk_app_config.remove_remote_ip" do |app|
+      app.middleware.delete ActionDispatch::RemoteIp
+    end
   end
 end
--- a/lib/govuk_app_config/version.rb
+++ b/lib/govuk_app_config/version.rb
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "2.7.0".freeze
+  VERSION = "2.8.0".freeze
 end


benthorner:
Tested and works with Content Publisher ✅.

kevindew:
Thanks for digging into this. I think beyond signon this would also affect all apps log files wouldn't it? isn't this the mechanism to determine the ip? I'm a bit concerned that it's a bit heavy handed to remove this entirely as all our apps then lose the ability to easily know a resolved client ip.

Could you elaborate on it being unreliable? My understanding was that the problem likely didn't lie in Rails but that the CDN would forward whatever the user claimed was their IP - rather than using the IP that actually connected to the CDN.

benthorner:
Oops, I just realised I failed to link to the justification in the commit! I can update once we reach a conclusion. I started out assuming there was a configuration issue, but now I'm not so sure. I'm open to keeping it, but thought it best to start with [the recommendation in the docs for this middleware](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html), which link to [a much longer article](https://blog.gingerlime.com/2012/rails-ip-spoofing-vulnerabilities-and-protection/) about the two headers at play, and recommends to remove the middleware. I also found [the code comments for the check](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L120-L132) quite helpful. 

My summary:

- The middleware is a known imperfect solution to derive the "true" client IP.
- The middleware sets `remote_ip`, which we basically don't use anywhere.
- We should avoid using it anywhere if possible, since we know it's flawed.

I guess an alternative would be to wipe the "Client-IP" header at the CDN level, which is similar to what you suggested. But as I understand it, the current behaviour of the header is correct - it's just that our chain of proxies subsequently adds "X-Forwarded-For", which the middleware doesn't like because it can only cope with one audit header. We could add the value of the "Client-IP" header into "X-Forwarded-For" before we wipe it, which would be more correct.

> I think beyond signon this would also affect all apps log files wouldn't it?

As I understand it, most of our request logging comes from Nginx, which is unaffected by this change.

Does that help?

kevindew:
> Does that help?

Yes thank you 👍 

> Oops, I just realised I failed to link to the justification in the commit! I can update once we reach a conclusion. I started out assuming there was a configuration issue, but now I'm not so sure. I'm open to keeping it, but thought it best to start with [the recommendation in the docs for this middleware](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html), which link to [a much longer article](https://blog.gingerlime.com/2012/rails-ip-spoofing-vulnerabilities-and-protection/) about the two headers at play, and recommends to remove the middleware. I also found [the code comments for the check](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L120-L132) quite helpful.

Yup, I can understand their concern with: "IF YOU DON'T USE A PROXY, THIS MAKES YOU VULNERABLE TO IP SPOOFING" It just surprises me as we use (rather a lot) of proxies which makes me wonder if we're just configured in an unusual way that means we get this.

> 
> My summary:
> 
> * The middleware is a known imperfect solution to derive the "true" client IP.
> * The middleware sets `remote_ip`, which we basically don't use anywhere.
> * We should avoid using it anywhere if possible, since we know it's flawed.
> 
> I guess an alternative would be to wipe the "Client-IP" header at the CDN level, which is similar to what you suggested. But as I understand it, the current behaviour of the header is correct - it's just that our chain of proxies subsequently adds "X-Forwarded-For", which the middleware doesn't like because it can only cope with one audit header. We could add the value of the "Client-IP" header into "X-Forwarded-For" before we wipe it, which would be more correct.

Gotcha -  the imperfect aspect seems to be that the Rails app needs to trust that at the first trusted proxy server there aren't any tampering with the X-FORWARDED-FOR or CLIENT-IP headers. I don't think it's more or less flawed than any other IP systems - they all have to rely on something that is trusted in the chain.

My understanding is that we seem to accept whatever a user sets at the CDN level when it should be nuked and reset since that is the edge of our network. For example, since you're only [supposed to have one of Client-Ip or X-Forwarded-For](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L120-L121) we should probably delete any entered Client-Ip header (or reset it, given it's not trusted) and [reset any given X-Forwarded-For header to be just the true client ip](https://docs.fastly.com/en/guides/manipulating-the-x-forwarded-for-header) - though this last step is only necessary if we don't want any untrusted input in logs.

Trying some experiments I saw:

If I send `curl -I -H "Client-Ip: 5.5.5.5" https://www.gov.uk/email-signup\?topic\=/transport/aviation-passenger-experience\&client-ip\=5.5.5.5`

Rails raises an error and CDN returns a 503
```
IP spoofing attack?! HTTP_CLIENT_IP="5.5.5.5" HTTP_X_FORWARDED_FOR="213.86.153.214, 157.52.92.41, 213.86.153.214, 34.246.209.74, 213.86.153.214, 127.0.0.1, 127.0.0.1, 10.12.4.102, 10.12.5.147"
```

Where the problem is that it's accepting the user provided client-ip.

If I send: `curl -I -H "X-Forwarded-For: 5.5.5.5" https://www.gov.uk/email-signup\?topic\=/transport/aviation-passenger-experience\&x-forwarded-for\=5.5.5.5` then I get a 200 response

Nginx logs flag the request as having come through this IP I spoofed: `5.5.5.5, 213.86.153.214, 185.31.16.80, 213.86.153.214, 127.0.0.1, 127.0.0.1, 10.13.5.15` - but Rails logs the IP correct as 213.86.153.214 (as it is the first non local subnet one if you work through the nginx list in reverse order.

And then if I do `curl -I -H "Client-Ip: 5.5.5.5" -H "X-Forwarded-For: 5.5.5.5" https://www.gov.uk/email-signup\?topic\=/transport/aviation-passenger-experience\&x-forwarded-for\=5.5.5.5\&client-ip\=5.5.5.5`

We get the same experience as the above with x-forwarded-for. Nginx logs has x-forwarded-for as `5.5.5.5, 213.86.153.214, 185.31.16.80, 213.86.153.214, 127.0.0.1, 127.0.0.1, 10.13.5.15` and Rails has the correct IP of 213.86.153.214. It doesn't throw an exception because the client-ip given (5.5.5.5) is listed in the x-forwarded-for header.

> > I think beyond signon this would also affect all apps log files wouldn't it?
> 
> As I understand it, most of our request logging comes from Nginx, which is unaffected by this change.
> 

The logs in Kibana that are just the app name (e.g whitehall rather than whitehall-admin.json.event.access) come from Rails and I'd suggest they're more commonly used than the nginx equivalents. Nginx doesn't actually have a particular field for ip (just x-forwarded-for) so the Rails logs appear the simpler way to query by origin ip (outside of Athena, which is best option).

benthorner:
Thanks for digging into this @kevindew! I think we're coming from slightly different perspectives, but that may be because I have less experience (and thus appreciation) of the use of IPs in our logs. If we're going to make use of these headers, then I agree it's necessary to ensure they make some (more) sense (than they do now).

> so the Rails logs appear the simpler way to query by origin ip

Fair enough.

To that end, I've made an alternative PR to sanitise them at the CDN level: https://github.com/alphagov/govuk-cdn-config/pull/302.

> Nginx logs flag the request as having come through this IP I spoofed: 5.5.5.5, 213.86.153.214, 185.31.16.80, 213.86.153.214, 127.0.0.1, 127.0.0.1, 10.13.5.15

I hadn't looked at the specific IPs before, and as discussed, it's surprising to see the VPN gateway IP duplicated. I didn't see that when I was looking at the logs on the `cache` machines. I did wonder if it could be that the request is going through the VPN gateway twice, when it passes between machines, but that's just a guess. Fortunately I don't think we need to delve into that hole as part of looking into this issue.

PR #181, opened 2021-01-07
Use statsd-ruby 1.5 client

bilbof:
This partially reverts https://github.com/alphagov/govuk_app_config/pull/180 and starts using the latest version ([1.5](https://rubygems.org/gems/statsd-ruby)) of the statsd client, which fixes TCP support.

The statsd client we were using was a temporary replacement for statsd-ruby - see the other PR for details.


richardTowers (approved):
Nice job getting this upstreamed!

PR #187, opened 2021-03-18
Adds plural rules not available in i18n

MuriloDalRi:
We have these rules in four different apps so moving them to one place
to make it easier to maintain.

https://trello.com/c/eEphXlio/2394-5-check-and-update-rules-for-plural-forms-across-frontend-apps

### DIFF
--- a/lib/govuk_app_config/govuk_i18n.rb
+++ b/lib/govuk_app_config/govuk_i18n.rb
@@ -1,0 +1,38 @@
+module GovukI18n
+  def plurals
+    {
+      # Dari - this isn't an iso code. Probably should be 'prs' as per ISO 639-3.
+      dr: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+      # Latin America and Caribbean Spanish
+      "es-419": { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+      # Scottish Gaelic
+      gd: { i18n: { plural: { keys: %i[one two few other],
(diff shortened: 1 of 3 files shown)


lib/govuk_app_config/govuk_i18n.rb
```diff
@@ -0,0 +1,36 @@
+module GovukI18n
+  {
```
thomasleese, line 2: I wonder if we should give this an explicit name such as

```rb
def plurals
  ...
end
```

or 

```rb
PLURALS = ...
```

Otherwise I don't think it'll be possible to reference this hash from another location.

lib/govuk_app_config/govuk_i18n.rb
```diff
@@ -0,0 +1,36 @@
+module GovukI18n
```
brucebolt, line 1: Could the name be changed so this is clearer that it relates to only the pluralisation rules, maybe something like `GovukI18nPlurals`?  Since this only returns a hash, it can't be expanded to any other I18n uses, so the name could potentially be misleading.

lib/govuk_app_config/govuk_i18n.rb
MuriloDalRi: Yeah, I'll turn that into a method

lib/govuk_app_config/govuk_i18n.rb
MuriloDalRi: I think I'll turn it into a `plurals` method so we can expand this in the future

brucebolt (approved):
Looks great, thanks for making the change

PR #196, opened 2021-05-06
Merge should_capture into before_send

benjamineskola:
https://trello.com/c/13nqv4jy/2474-3-merge-govukerrors-shouldcapture-method-into-beforesend-🍐

### DIFF
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -12,32 +12,26 @@ module GovukError
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
       self.active_sentry_environments = []
       self.data_sync_excluded_exceptions = []
-      self.should_capture = ignore_exceptions_based_on_env_and_data_sync
+      @before_send_callbacks = [
+        ignore_exceptions_if_not_in_active_sentry_env,
+        ignore_excluded_exceptions_in_data_sync,
+        increment_govuk_statsd_counters,
+      ]
+      super.before_send = run_before_send_callbacks
     end
 
-    def should_capture=(closure)
-      combined = lambda do |error_or_event|
-        (ignore_exceptions_based_on_env_and_data_sync.call(error_or_event) && closure.call(error_or_event))
-      end
-
-      super(combined)
+    def before_send=(closure)
+      @before_send_callbacks.insert(-2, closure)
     end
 
   protected
 
-    def ignore_exceptions_based_on_env_and_data_sync
-      lambda do |error_or_event|
-        ignore_exceptions_if_not_in_active_sentry_env.call(error_or_event) &&
-          ignore_excluded_exceptions_in_data_sync.call(error_or_event)
-      end
-    end
-
     def ignore_exceptions_if_not_in_active_sentry_env
-      ->(_error_or_event) { active_sentry_environments.include?(sentry_environment) }
+      ->(error_or_event, _hint) { error_or_event if active_sentry_environments.include?(sentry_environment) }
     end
 
     def ignore_excluded_exceptions_in_data_sync
-      lambda { |error_or_event|
+      lambda { |error_or_event, _hint|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
           exception_to_ignore = Object.const_get(exception_to_ignore) unless exception_to_ignore.is_a?(Module)
           exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
@@ -45,11 +39,30 @@ module GovukError
         rescue NameError
           # the exception type represented by the exception_to_ignore string
           # doesn't even exist in this environment, so won't be found in the chain
-          false
+          nil
         end
 
-        !(data_sync.in_progress? && data_sync_ignored_error)
+        error_or_event unless data_sync.in_progress? && data_sync_ignored_error
       }
+    end
+
+    def increment_govuk_statsd_counters
+      lambda { |error_or_event, _hint|
+        GovukStatsd.increment("errors_occurred")
+        GovukStatsd.increment("error_types.#{error_or_event.class.name.demodulize.underscore}")
+        error_or_event
+      }
+    end
+
+    def run_before_send_callbacks
+      lambda do |error_or_event, hint|
+        result = error_or_event
+        @before_send_callbacks.each do |callback|
+          result = callback.call(error_or_event, hint)
+          break if result.nil?
+        end
+        result
+      end
     end
   end
 end
--- a/lib/govuk_app_config/govuk_error/configure.rb
+++ b/lib/govuk_app_config/govuk_error/configure.rb
@@ -1,10 +1,4 @@
 GovukError.configure do |config|
-  config.before_send = proc { |e|
-    GovukStatsd.increment("errors_occurred")
-    GovukStatsd.increment("error_types.#{e.class.name.demodulize.underscore}")
-    e
-  }
-
   config.silence_ready = !Rails.env.production? if defined?(Rails)
 
   # These are the environments (described by the `SENTRY_CURRENT_ENV`
--- a/spec/govuk_error/configuration_spec.rb
+++ b/spec/govuk_error/configuration_spec.rb
@@ -11,20 +11,20 @@ RSpec.describe GovukError::Configuration do
     end
   end
 
-  describe ".should_capture" do
+  describe ".before_send" do
     let(:configuration) { GovukError::Configuration.new(Raven.configuration) }
 
     it "ignores errors if they happen in an environment we don't care about" do
       ClimateControl.modify SENTRY_CURRENT_ENV: "some-temporary-environment" do
         configuration.active_sentry_environments << "production"
-        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+        expect(configuration.before_send.call(StandardError.new)).to be_nil
       end
     end
 
     it "captures errors if they happen in an environment we care about" do
       ClimateControl.modify SENTRY_CURRENT_ENV: "production" do
         configuration.active_sentry_environments << "production"
-        expect(configuration.should_capture.call(StandardError.new)).to eq(true)
+        expect(configuration.before_send.call(StandardError.new)).to be_truthy
       end
     end
 
@@ -39,19 +39,19 @@ RSpec.describe GovukError::Configuration do
       end
 
       it "should capture errors by default" do
-        expect(configuration.should_capture.call(StandardError.new)).to eq(true)
+        expect(configuration.before_send.call(StandardError.new)).to be_truthy
       end
 
       it "should ignore errors that have been added as a string to data_sync_excluded_exceptions" do
         configuration.data_sync_excluded_exceptions << "StandardError"
 
-        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+        expect(configuration.before_send.call(StandardError.new)).to be_nil
       end
 
       it "should ignore errors that have been added as a class to data_sync_excluded_exceptions" do
         configuration.data_sync_excluded_exceptions << StandardError
 
-        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+        expect(configuration.before_send.call(StandardError.new)).to be_nil
       end
 
       it "should ignore errors whose underlying cause is an exception in data_sync_excluded_exceptions" do
@@ -71,7 +71,7 @@ RSpec.describe GovukError::Configuration do
         end
 
         expect(chained_exception).to be_an_instance_of(SomeOtherError)
-        expect(configuration.should_capture.call(chained_exception)).to eq(false)
+        expect(configuration.before_send.call(chained_exception)).to be_nil
       end
 
       it "should ignore errors that are subclasses of an exception in data_sync_excluded_exceptions" do
@@ -79,7 +79,7 @@ RSpec.describe GovukError::Configuration do
         stub_const("SomeInheritedClass", Class.new(SomeClass))
 
         configuration.data_sync_excluded_exceptions << "SomeClass"
-        expect(configuration.should_capture.call(SomeInheritedClass.new)).to eq(false)
+        expect(configuration.before_send.call(SomeInheritedClass.new)).to be_nil
       end
     end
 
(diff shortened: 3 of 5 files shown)


thomasleese (commented):
Code looks fine, I think we should add more reasoning to the commit message about why we're making this change, and add an entry to the CHANGELOG.

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -12,12 +12,14 @@ def initialize(_raven_configuration)
...
+        (ignore_exceptions_based_on_env_and_data_sync.call(error_or_event, hint) && increment_govuk_statsd_counters.call(error_or_event, hint))
+      end
     end
 
-    def should_capture=(closure)
-      combined = lambda do |error_or_event|
-        (ignore_exceptions_based_on_env_and_data_sync.call(error_or_event) && closure.call(error_or_event))
+    def before_send=(closure)
+      combined = lambda do |error_or_event, hint|
+        (ignore_exceptions_based_on_env_and_data_sync.call(error_or_event, hint) && increment_govuk_statsd_counters.call(error_or_event, hint) && closure.call(error_or_event, hint))
```
ChrisBAshton, line 22: I wonder if it's worth extracting the 'default' behaviour into its own method, so that we remove the duplication between lines 16 and 22. E.g.

```ruby
def before_send=(closure)
  combined = lambda do |error_or_event, hint|
    default_before_send_actions.call(error_or_event, hint) && closure.call(error_or_event, hint)
  end
  super(combined)
end
```

I wonder what happens now if we call `before_send=` 🤔  I can't remember if it stacks on top of previous calls, e.g.

```ruby
before_send = do_something
before_send = do_something_else

# when an error DOES happen, what then gets evaluated?
# is it:
[ default_before_send_actions, do_something, do_something_else ]
# or is it:
[ default_before_send_actions, do_something, default_before_send_actions, do_something, do_something_else ]
```

I hope it's the former, but if it's the latter, it hasn't been a problem until now, because it just evaluates whether or not to log the error. But now that _increments_ are being brought into the mix, we want to make sure that we're not double-counting. Does that make sense?
Would it be possible to add a test to check that behaviour?

lib/govuk_app_config/govuk_error/configure.rb
```diff
@@ -1,10 +1,4 @@
 GovukError.configure do |config|
-  config.before_send = proc { |e|
-    GovukStatsd.increment("errors_occurred")
```
ChrisBAshton, line 3: Looks like this never had a test, which is unfortunate. Would it be possible to add one now that we've moved the code? Would be great to have a test that confirms that these statsd values are incremented when an error occurs.

spec/govuk_error/configuration_spec.rb
```diff
@@ -11,20 +11,20 @@
...
 
-  describe ".should_capture" do
+  describe ".before_send" do
     let(:configuration) { GovukError::Configuration.new(Raven.configuration) }
 
     it "ignores errors if they happen in an environment we don't care about" do
       ClimateControl.modify SENTRY_CURRENT_ENV: "some-temporary-environment" do
         configuration.active_sentry_environments << "production"
-        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+        expect(configuration.before_send.call(StandardError.new)).to be_falsy
```
ChrisBAshton, line 20: Any particular reason why we've dropped `eq(false)` to `be_falsy`? Explicit boolean checks tend to be better, as `be_falsy` will pass even if the returned value is 0, which could be surprising.

ChrisBAshton (commented):
Thanks Ben - it's shaping up nicely. Just had a few questions/suggestions before we merge:

spec/govuk_error/configuration_spec.rb
benjamineskola: I made this change because the behaviour of `should_capture` is to return a boolean specifically, but `before_send` ordinarily just returns its input, but returns `nil` if it's to be dropped.

That was just the lazy way of testing it, though — I could also change these to look for `nil` or the input explicitly.

lib/govuk_app_config/govuk_error/configuration.rb
benjamineskola: Good point — I don't _think_ it would stack (which is why the default is set in the constructor and then prepended in the setter method), but it can't hurt to have a test to ensure this.

benjamineskola: Looks like this is what's happening: because setting `before_send` in `initalize` ends up setting it to `default_before_send_actions && default_before_send_actions`.

[This](https://github.com/alphagov/govuk_app_config/pull/196/files#diff-59d0fe03fd43214b7f65eb27094e958f04c10366cbf2c5b5e3e470483899fc72R15) is a bit of a hack, but on the other hand, if `before_send` isn't initialised at all then the default actions won't be set either. I'm not sure if there's a better solution; I wondered about having an array of lambdas which `before_send` could call in turn, and which clients could add to if necessary, but that would mean altering the interface for dependants (and setting `before_send` directly would break it).

spec/govuk_error/configuration_spec.rb
```diff
@@ -11,20 +11,20 @@
...
 
-  describe ".should_capture" do
+  describe ".before_send" do
     let(:configuration) { GovukError::Configuration.new(Raven.configuration) }
 
     it "ignores errors if they happen in an environment we don't care about" do
       ClimateControl.modify SENTRY_CURRENT_ENV: "some-temporary-environment" do
         configuration.active_sentry_environments << "production"
-        expect(configuration.should_capture.call(StandardError.new)).to eq(false)
+        expect(configuration.before_send.call(StandardError.new)).to be_nil
```
ChrisBAshton, line 20: Ah, I see - because it needs to return `nil` to drop the event: https://docs.sentry.io/platforms/ruby/configuration/filtering/

Could you explain that in the commit message when you come to rebase this?

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -51,5 +52,19 @@ def ignore_excluded_exceptions_in_data_sync
         !(data_sync.in_progress? && data_sync_ignored_error)
       }
     end
+
+    def increment_govuk_statsd_counters
+      lambda { |error_or_event, _hint|
+        GovukStatsd.increment("errors_occurred")
+        GovukStatsd.increment("error_types.#{error_or_event.class.name.demodulize.underscore}")
```
ChrisBAshton, line 59: Argh - sorry to throw another spanner in the works, but I've just realised we wouldn't want to increment these values unless we actually end up logging the error. As it stands, we could set our own `before_send=` lambda to check an error type and decide to return `nil`, but by that point we've already evaluated `increment_govuk_statsd_counters` and thus would have incorrectly incremented our global error count.

This can be in its own separate commit: could you write a test that ensures that `increment_govuk_statsd_counters` is the last callback that gets called? I'm _hoping_ it's actually not too much work - should just be a case of reversing 

`default_before_send_actions.call(error_or_event, hint) && closure.call(error_or_event, hint)` 
=>
`closure.call(error_or_event, hint) && default_before_send_actions.call(error_or_event, hint)`

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -12,12 +12,13 @@ def initialize(_raven_configuration)
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
       self.active_sentry_environments = []
       self.data_sync_excluded_exceptions = []
-      self.should_capture = ignore_exceptions_based_on_env_and_data_sync
+      self.before_send = ->(error_or_event, _hint) { error_or_event }
```
ChrisBAshton, line 15: Hmmm, if I've understood https://github.com/alphagov/govuk_app_config/pull/196#discussion_r629452794 then we could avoid the stubbed lambda by swapping the above with this:

```ruby
      self.before_send = default_before_send_actions
```

...and then the method would be:

```ruby
def before_send=(closure)
  super(closure)
end
```

That does seem deceptively simple though, so I suspect it would be overriding previous lambdas rather than merging them. Perhaps one to consider refactoring after writing the "two `before_send=` test" suggested above.

benjamineskola: Yeah, because it's overriding then the `before_send=` method needs to do the work of merging if a non-default one is chosen; however, that also applies when the default one is being set and leads to doubling the default — if and only if a non-default one is not set (hence this stub).

What would be easiest (and what I assumed at first) would be if `before_send=` is only defined _after_ the `initialize` is called; or otherwise, if the method could check whether the default is being set. But that ends up needing to do some weird reflection or state (a `before_send_was_overriden` instance variable or something?).

lib/govuk_app_config/govuk_error/configuration.rb
benjamineskola: Hm, I think this needs some caution. If the user-defined callback returns nil then the default callback won't be run — but returning nil seems fairly common and innocuous (e.g., `GovukStatsd.increment('…')` returns nil which is how I realised).

The documentation for `before_send` says that it ought to return the input if all is well or nil if not, so it's something that _should_ be accounted for when writing the callback, but it also seems like it would be easy not to notice that the counters were no longer being incremented. Perhaps I'm overthinking it though.

spec/govuk_error/configuration_spec.rb
```diff
@@ -133,6 +133,36 @@ def test_closure
         end
       end
     end
+
+    context "when the before_send lambda has been overridden twice, the first does not take effect" do
```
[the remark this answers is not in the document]
ChrisBAshton: That sounds like a good approach - so long as we can keep the interface the same (i.e. consumers just call `before_send=`) then I don't think it matters much what our underlying implementation is.
Hit the re-review button when you're happy for me to take another look 👍

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: Yes, we may want to make it explicit in documentation somewhere, though there's only one instance of `before_send` being used on GOV.UK and they've [explicitly returned the error](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/37fd197b2d94e55f3ed759b381181d282c4a18c8/config/application.rb#L33), so I'm hoping this is a non-problem :)

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: Yes, we may want to make it explicit in documentation somewhere, though there's only one instance of `before_send` being used on GOV.UK and they've [explicitly returned the error](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/37fd197b2d94e55f3ed759b381181d282c4a18c8/config/application.rb#L33), so I'm hoping this is a non-problem :)

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -12,7 +12,7 @@ def initialize(_raven_configuration)
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
       self.active_sentry_environments = []
       self.data_sync_excluded_exceptions = []
-      self.before_send = ->(error_or_event, _hint) { error_or_event }
+      super.before_send = default_before_send_actions
```
ChrisBAshton: Genius! 👏

ChrisBAshton (commented):
Excellent work, @benjamineskola - the approach looks good and I think we have enough test coverage to be confident it works 🎉 

I think this is now ready for a rebase/tidy-up before a final review.

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -12,32 +12,32 @@ def initialize(_raven_configuration)
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
       self.active_sentry_environments = []
       self.data_sync_excluded_exceptions = []
-      self.should_capture = ignore_exceptions_based_on_env_and_data_sync
+      @before_send_callbacks = [
+        ignore_exceptions_based_on_env_and_data_sync,
```
ChrisBAshton, line 16: Now that this is an array, I don't think there's value in having a separate `ignore_exceptions_based_on_env_and_data_sync` method. I think we can just pass `ignore_exceptions_if_not_in_active_sentry_env` and `ignore_excluded_exceptions_in_data_sync` in here directly.

spec/govuk_error/configuration_spec.rb
ChrisBAshton: Minor: Might be worth fleshing out the commit message slightly by linking to the URL above ^

spec/govuk_error/configuration_spec.rb
```diff
@@ -96,22 +96,92 @@
...
+          configuration.before_send.call(StandardError.new)
+        end
       end
     end
   end
 
-  describe ".should_capture=" do
-    it "Allows apps to add their own `should_capture` callback, that is evaluated alongside the default. If both return `true`, then we should capture, but if either returns `false`, then we shouldn't." do
+  describe ".before_send=" do
+    it "Allows apps to add their own `before_send` callback, that is evaluated alongside the default. If both return `true`, then we should capture, but if either returns `false`, then we shouldn't." do
```
ChrisBAshton, line 160: This test description needs updating to refer to returning the `error_or_event` or `nil`

spec/govuk_error/configuration_spec.rb
```diff
@@ -96,22 +96,92 @@
...
+        raven_configurator = GovukError::Configuration.new(Raven.configuration)
+        raven_configurator.active_sentry_environments << "production"
+        raven_configurator.before_send = lambda do |_error_or_event, _hint|
+          nil
+        end
+
+        expect(GovukStatsd).not_to receive(:increment).with("errors_occurred")
+        expect(GovukStatsd).not_to receive(:increment).with("error_types.standard_error")
+
+        raven_configurator.before_send.call(StandardError.new)
```
ChrisBAshton, line 184: Loving these tests 👏

ChrisBAshton (commented):
Thanks Ben, sorry for not spotting that this had been rebased.

I've spotted a couple more minor things on final review, which I think would be worth incorporating before we merge 👍 it's looking really great now though.

ChrisBAshton (approved):
Great job! 👏

ChrisBAshton (approved):

PR #199, opened 2021-06-01
v4 pre-release: upgrade sentry-raven to sentry-ruby

ChrisBAshton:
On 10 December 2020, `sentry-raven` was [deprecated](https://github.com/getsentry/sentry-ruby/tree/master/sentry-raven) and renamed to [`sentry-ruby`](https://github.com/getsentry/sentry-ruby) (which is essentially `sentry-raven` "v4"). `sentry-raven` will [no longer receive any support or bug fixes](https://github.com/getsentry/sentry-ruby/blame/master/sentry-raven/README.md#L18).

The benefits of moving to `sentry-ruby` include:

- Unified interface with other SDKs (i.e. documentation that used to be irrelevant will now be relevant!)
- performance monitoring (though [this isn't included in our plan](https://sentry.io/organizations/govuk/performance/))
- "better extensibility... will allow the community to build extensions for different integrations/features".  (It's not clear what benefits that might bring just yet.)

This PR also fixes the `data_sync_excluded_exceptions` behaviour that has been broken since v3.1.0. See commits for details.

Trello: https://trello.com/c/1zVPYfTR/2539-3-replace-sentry-raven-with-sentry-ruby-in-govukappconfig

### DIFF
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -6,7 +6,7 @@ module GovukError
     attr_reader :data_sync, :sentry_environment
     attr_accessor :active_sentry_environments, :data_sync_excluded_exceptions
 
-    def initialize(_raven_configuration)
+    def initialize(_sentry_configuration)
       super
       @sentry_environment = ENV["SENTRY_CURRENT_ENV"]
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
@@ -34,7 +34,7 @@ module GovukError
       lambda { |error_or_event, _hint|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
           exception_to_ignore = Object.const_get(exception_to_ignore) unless exception_to_ignore.is_a?(Module)
-          exception_chain = Raven::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+          exception_chain = Sentry::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
           exception_chain.any? { |exception| exception.is_a?(exception_to_ignore) }
         rescue NameError
           # the exception type represented by the exception_to_ignore string
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -50,7 +50,7 @@ module GovukError
       lambda { |event, hint|
         if hint[:exception]
           GovukStatsd.increment("errors_occurred")
-          GovukStatsd.increment("error_types.#{hint[:exception].class.name.demodulize.underscore}")
+          GovukStatsd.increment("error_types.#{hint[:exception].class.name.split('::').last.underscore}")
         end
         event
       }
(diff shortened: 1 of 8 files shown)


thomasleese (approved):
I've had a brief look here https://docs.sentry.io/platforms/ruby/migration/ and it looks like it should all be compatible!

thomasleese (approved):

lib/govuk_app_config/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "3.1.1".freeze
+  VERSION = "4.0.0.pre-1".freeze
```
kevindew, line 2: FYI using a pre-1 name caused me a bunch of pain before: https://github.com/alphagov/rubocop-govuk/commit/be6bad2266dcce320beba08e3c831b5642c86d48. I'd recommend `pre.1`.

spec/govuk_error/configuration_spec.rb
```diff
@@ -27,14 +27,27 @@
...
     it "captures errors if they happen in an environment we care about" do
       ClimateControl.modify SENTRY_CURRENT_ENV: "production" do
         configuration.active_sentry_environments << "production"
-        expect(configuration.before_send.call(StandardError.new)).to be_truthy
+        sentry_client = send_exception_to_sentry(StandardError.new, configuration)
+        expect(sentry_client.transport.events.count).to eq(1)
+      end
+    end
+
+    it "allows string messages to be sent (rather than exceptions)" do
```
kevindew, line 43: Thanks for adding this test and your explanation on Slack that `Sentry::Utils::ExceptionCauseChain.exception_to_array()` can cope with an input of nil. One thing I noticed in trying that out was that this test doesn't execute https://github.com/alphagov/govuk_app_config/blob/2f377a28eb37983e14a4e4918e54b27fc8b9187e/lib/govuk_app_config/govuk_error/configuration.rb#L36-L42

since data_sync_excluded_exceptions is an empty array. I think it could be useful to have that covered in the test case as passing that method nil seems a risky integration point.

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -27,38 +27,38 @@ def before_send=(closure)
...
       }
     end
 
     def increment_govuk_statsd_counters
-      lambda { |error_or_event, _hint|
+      lambda { |event, hint|
         GovukStatsd.increment("errors_occurred")
-        GovukStatsd.increment("error_types.#{error_or_event.class.name.demodulize.underscore}")
-        error_or_event
+        GovukStatsd.increment("error_types.#{hint[:exception].class.name.split('::').last.underscore}")
```
kevindew, line 52: I think this will produce the rather cryptic `error_types.nil_class` if hint[:exception] is nil. Might be best to have this behind a guard clause.

kevindew (commented):
Thanks for all you've done on this Chris - seems great. It's not the most full review from me, just looking at the bits we were discussing on Slack.

lib/govuk_app_config/version.rb
ChrisBAshton: Thanks for the heads up.

spec/govuk_error/configuration_spec.rb
ChrisBAshton: Good spot. I've now populated the array.

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: I've added a test for this - you were right 👍

PR #200, opened 2021-06-01
Incorporate latest rubocop rules

ChrisBAshton:
Upgrading rubocop-govuk to version 4 (via running `bundle update` locally) introduces some new rules we need to account for.

We've had to specify a minimum Ruby version, which we have set as 2.6.5 in `.ruby-version`, so I've used that as the basis.
Then it's mostly warnings about declaring classes within tests. Where possible, I've swapped these out for `stub_const`, but we have several tests which rely heavily on the internals of class declaration, inheritence, method overloading and the like, and it seemed less expressive to try to convert these in a way that appeased the linter. I've therefore added some lint-ignore
declarations inline in these files.

### DIFF
--- a/lib/govuk_app_config/govuk_error/govuk_data_sync.rb
+++ b/lib/govuk_app_config/govuk_error/govuk_data_sync.rb
@@ -3,14 +3,8 @@ require "time"
 module GovukError
   class GovukDataSync
     class MalformedDataSyncPeriod < RuntimeError
-      attr_reader :invalid_value
-
       def initialize(invalid_value)
-        @invalid_value = invalid_value
-      end
-
-      def message
-        "\"#{invalid_value}\" is not a valid value (should be of form '22:00-03:00')."
+        super("\"#{invalid_value}\" is not a valid value (should be of form '22:00-03:00').")
       end
     end
 
--- a/spec/lib/govuk_content_security_policy_spec.rb
+++ b/spec/lib/govuk_content_security_policy_spec.rb
@@ -3,7 +3,11 @@ require "rails"
 require "govuk_app_config/govuk_content_security_policy"
 
 RSpec.describe GovukContentSecurityPolicy do
-  class DummyCspRailsApp < Rails::Application; end
+  before do
+    stub_const("DummyCspRailsApp", Class.new(Rails::Application))
+  end
+
+  after { Rails.application = nil }
 
   describe ".configure" do
     it "creates a policy" do
--- a/spec/lib/govuk_healthcheck/checkup_spec.rb
+++ b/spec/lib/govuk_healthcheck/checkup_spec.rb
@@ -2,11 +2,79 @@ require "spec_helper"
 require "govuk_app_config/govuk_healthcheck"
 
 RSpec.describe GovukHealthcheck::Checkup do
-  let(:ok_check) { OkTestHealthcheck }
-  let(:warning_check) { WarningTestHealthcheck }
-  let(:critical_check) { CriticalTestHealthcheck }
-  let(:disabled_critical_check) { DisabledCriticalHealthcheck }
-  let(:exception_check) { ExceptionHealthcheck }
+  let(:test_healthcheck) do
+    Class.new do
+      def name
+        "#{status}_check".to_sym
+      end
+
+      def status
+        :unknown
+      end
+    end
+  end
+
+  let(:ok_check) do
+    Class.new(test_healthcheck) do
+      def status
+        :ok
+      end
+    end
+  end
+
+  let(:ok_check_with_message) do
+    Class.new(ok_check) do
+      def message
+        "This is a custom message"
+      end
+    end
+  end
+
+  let(:ok_check_with_details) do
+    Class.new(ok_check) do
+      def details
+        {
+          extra: "This is an extra detail",
+        }
+      end
+    end
+  end
+
+  let(:warning_check) do
+    Class.new(test_healthcheck) do
+      def status
+        :warning
+      end
+    end
+  end
+
+  let(:critical_check) do
+    Class.new(test_healthcheck) do
+      def status
+        :critical
+      end
+    end
+  end
+
+  let(:disabled_critical_check) do
+    Class.new(critical_check) do
+      def enabled?
+        false
+      end
+    end
+  end
+
+  let(:exception_check) do
+    Class.new do
+      def name
+        :exception_check
+      end
+
+      def status
+        raise "something bad happened"
+      end
+    end
+  end
 
   it "sets the overall status to the worse component status" do
     expect(described_class.new([ok_check]).run[:status]).to eq(GovukHealthcheck::OK)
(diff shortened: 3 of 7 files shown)


ChrisBAshton:
Hmm, govuk_logging_spec currently fails via `bundle exec rspec --seed 62442` with:

```
     Failure/Error: expect(lines).to include(/default exception/)
     
       expected ["/Users/christopherashton/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/logstasher-2.1.5/lib/logsta...ils_ext/action_dispatch/debug_exceptions.rb:32: warning: previous definition of log_error was here"] to include /default exception/
       Diff:
       @@ -1,6 +1,11 @@
       -[/default exception/]
       +["/Users/christopherashton/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/logstasher-2.1.5/lib/logstasher.rb:74: warning: method redefined; discarding old logstasher_add_custom_fields_to_payload",
       + "/Users/christopherashton/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/logstasher-2.1.5/lib/logstasher.rb:70: warning: previous definition of logstasher_add_custom_fields_to_payload was here",
       + "/Users/christopherashton/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/logstasher-2.1.5/lib/logstasher.rb:75: warning: method redefined; discarding old logstasher_add_custom_fields_to_payload",
       + "/Users/christopherashton/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/logstasher-2.1.5/lib/logstasher.rb:70: warning: previous definition of logstasher_add_custom_fields_to_payload was here",
       + "/Users/christopherashton/govuk/govuk_app_config/lib/govuk_app_config/rails_ext/action_dispatch/debug_exceptions.rb:32: warning: method redefined; discarding old log_error",
       + "/Users/christopherashton/govuk/govuk_app_config/lib/govuk_app_config/rails_ext/action_dispatch/debug_exceptions.rb:32: warning: previous definition of log_error was here"]
```

thomasleese:
I've amended the PR, and the tests are passing now!

spec/lib/govuk_content_security_policy_spec.rb
```diff
@@ -2,9 +2,9 @@
 require "rails"
 require "govuk_app_config/govuk_content_security_policy"
 
-RSpec.describe GovukContentSecurityPolicy do
-  class DummyCspRailsApp < Rails::Application; end
+class DummyCspRailsApp < Rails::Application; end
```
kevindew, line 5: Rather than defining a global constant, this could be set-up scoped to the test with:

```
  before do
    stub_const("DummyCspRailsApp", Class.new(Rails::Application))
  end
```

spec/lib/govuk_healthcheck/checkup_spec.rb
```diff
@@ -1,6 +1,64 @@
...
+
+class ExceptionHealthcheck
+  def name
+    :exception_check
+  end
+
+  def status
+    raise "something bad happened"
+  end
+end
```
kevindew, line 60: Rather than defining these classes in the global namespace I think we can create them with lets (since I think they only need to be accessed locally). For example:

```
  let(:test_healthcheck) do
    Class.new do
      def name
        "#{status}_check".to_sym
      end

      def status
        :unknown
      end
    end
  end

  let(:ok_check) do
    Class.new(test_healthcheck) do
      def status
        :ok
      end
    end
  end

  let(:warning_check) do
    Class.new(test_healthcheck) do
      def status
        :warning
      end
    end
  end

  let(:critical_check) do
    Class.new(test_healthcheck) do
      def status
        :critical
      end
    end
  end

  let(:disabled_critical_check) do
    Class.new(critical_check) do
      def enabled?
        false
      end
    end
  end

  let(:exception_check) do
    Class.new do
      def name
        :exception_check
      end

      def status
        raise "something bad happened"
      end
    end
  end
```

spec/lib/rails_ext/action_dispatch/debug_exceptions_spec.rb
```diff
@@ -2,6 +2,24 @@
...
+
+class RightParametersTestClass
+private
+
+  def log_error(request, wrapper); end
+end
+
+class FakeDebugExceptions
+  def log_error(request, wrapper); end
+end
```
kevindew, line 21: I think these classes can be defined by lets too

spec/lib/govuk_logging_spec.rb
```diff
@@ -3,14 +3,14 @@
...
-    config.hosts.clear
-    routes.draw do
-      get "/error", to: proc { |_env| raise Exception, "default exception" }
-    end
+class DummyLoggingRailsApp < Rails::Application
+  config.hosts.clear
+  routes.draw do
+    get "/error", to: proc { |_env| raise StandardError, "default exception" }
   end
+end
```
kevindew, line 11: I think this would be better suited to a stub_const within the test

kevindew (commented):
Thanks for your work on this. I've added a bunch of suggestions around the class configurations to avoid leaking constants globally.

spec/lib/govuk_content_security_policy_spec.rb
thomasleese: This is actually what we'd done in the first attempt, but it causes issues where the `Rails::Application` wasn't being "registered" properly and tests that relied on that were failing... I can try and investigate further but switching away from `stub_const` fixed this.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
```diff
@@ -6,6 +6,7 @@ class MalformedDataSyncPeriod < RuntimeError
       attr_reader :invalid_value
 
       def initialize(invalid_value)
+        super
```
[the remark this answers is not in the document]
thomasleese: Ah yes! I thought `invalid_value` was being used elsewhere, but it's not.

Also I'm surprised to learn that calling `super` on its own automatically passes the arguments through to the super class... I would've expected it to be explicitly required.

spec/lib/rails_ext/action_dispatch/debug_exceptions_spec.rb
thomasleese: Good point, I've fixed that in e8a04dd6be5854056dfb7bf4a3cb259fdeb8142a.

spec/lib/govuk_healthcheck/checkup_spec.rb
thomasleese: Ah yep, I've fixed that in c81a6af.

kevindew: Nice one - just a heads up that you've still got the old classes defined.

lib/govuk_app_config/govuk_error/govuk_data_sync.rb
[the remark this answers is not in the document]
kevindew: Yeah, it's a bit of a surprising quirk for Ruby 😞 . You have to do `super()` to send it with no args which tends to look rather wrong.

spec/lib/govuk_content_security_policy_spec.rb
kevindew: Ah ha sorry about that, I gave it a quick spin before commenting and didn't experience a test failure. It'll be fine if we do have to do it with a real class but it'd be good to document why.

thomasleese: I was able to get it to work, but I had to make an extra change (which I believe was only working by accident before). I've explained in c956a6d.

kevindew: Great stuff. I think I've worked out what the problem is. As `Rails.application` is a class variable it ends up holding on to state in between tests. I think you can avoid the fix you've put in by adding an after block of `after { Rails.application = nil }` in both of the specs that use a Rails::Application, which can reset it after execution.

thomasleese: Ooh nice, that's cleaner, I've amended the commit to do that!

kevindew (approved):
Awesome 👍

PR #201, opened 2021-06-03
Query the original exception rather than Sentry::Event object

ChrisBAshton:
In the `before_send` callbacks, the first parameter is of type `Sentry::Event`, and the second parameter (`hint`) is a hash which contains the original exception at key `exception`.

Our datasync-ignorable-exception logic relies on querying the exception chain to see if any of the exceptions in the chain are
of an ignorable exception type. This logic doesn't work when we view the `Sentry::Event` object (as we have been doing) - we must instead look at the `hint[:exception]` object.

The consequence of this bug was that our logic to ignore certain exceptions that occur during the datasync, no longer worked, and we've therefore seen many thousands of Sentry errors since #196 was merged.

However, it wasn't as simple as just checking the `hint` parameter. The `hint` parameter was missing, due to the way we had constructed our tests. Sentry automatically provides the `hint` parameter, but only if the event has been sent through all of the layers of Sentry's architecture, which took a bit of trial and error to get right. I took inspiration from [hub_spec.rb] and [sentry_spec.rb] from the Sentry gem repository.

This has also exposed how our GovukStatsd incrementing was not working, as it would log all errors as `error_types.sentry_event` rather than their original exception types. This is fixed too.

[hub_spec.rb]: https://github.com/getsentry/sentry-ruby/blob/e4b6d66612d9ce601596c23a5dd0d035c559d980/sentry-ruby/spec/sentry/hub_spec.rb
[sentry_spec.rb]: https://github.com/getsentry/sentry-ruby/blob/ace176cfceaefc91e29364a1c4180a67337e8ab0/sentry-ruby/spec/sentry_spec.rb

### DIFF
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -27,14 +27,14 @@ module GovukError
   protected
 
     def ignore_exceptions_if_not_in_active_sentry_env
-      ->(error_or_event, _hint) { error_or_event if active_sentry_environments.include?(sentry_environment) }
+      ->(event, _hint) { event if active_sentry_environments.include?(sentry_environment) }
     end
 
     def ignore_excluded_exceptions_in_data_sync
-      lambda { |error_or_event, _hint|
+      lambda { |event, hint|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
           exception_to_ignore = Object.const_get(exception_to_ignore) unless exception_to_ignore.is_a?(Module)
-          exception_chain = Sentry::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+          exception_chain = Sentry::Utils::ExceptionCauseChain.exception_to_array(hint[:exception])
           exception_chain.any? { |exception| exception.is_a?(exception_to_ignore) }
         rescue NameError
           # the exception type represented by the exception_to_ignore string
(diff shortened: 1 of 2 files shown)


thomasleese (approved):

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -27,38 +27,38 @@ def before_send=(closure)
...
+      ->(event, _hint) { event if active_sentry_environments.include?(sentry_environment) }
     end
 
     def ignore_excluded_exceptions_in_data_sync
-      lambda { |error_or_event, _hint|
+      lambda { |event, hint|
         data_sync_ignored_error = data_sync_excluded_exceptions.any? do |exception_to_ignore|
           exception_to_ignore = Object.const_get(exception_to_ignore) unless exception_to_ignore.is_a?(Module)
-          exception_chain = Sentry::Utils::ExceptionCauseChain.exception_to_array(error_or_event)
+          exception_chain = Sentry::Utils::ExceptionCauseChain.exception_to_array(hint[:exception])
```
kevindew, line 37: Are we prepared for this to be nil? I assume this isn't set if you do: `Sentry.capture_message("hello world!")`

lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -27,38 +27,38 @@ def before_send=(closure)
...
       }
     end
 
     def increment_govuk_statsd_counters
-      lambda { |error_or_event, _hint|
+      lambda { |event, hint|
         GovukStatsd.increment("errors_occurred")
-        GovukStatsd.increment("error_types.#{error_or_event.class.name.demodulize.underscore}")
-        error_or_event
+        GovukStatsd.increment("error_types.#{hint[:exception].class.name.demodulize.underscore}")
```
kevindew, line 52: Similar comment about nil - we could have error_types.nil_class come through here. I'm not sure what came through in the past if someone used `capture_message` maybe we have some error_types.sentry_event?

By the way, small aside that has probably not been noticed or caused a problem in the past. The `demodularize` method is part of ActiveSupport which isn't a dependency of this Repo. It's a [very simple method](https://github.com/rails/rails/blob/d3f4db9e95a475822c02b29241a5d07cbcff6fd5/activesupport/lib/active_support/inflector/methods.rb#L228-L235) so we can probably just replace it with some inline code rather than requiring activesupport (`"::Test".split("::").last`)

kevindew (commented):
Thanks Chris - great to have this sorted. I've added comments on a couple of things I think you may have missed - apologies if already in hand.

PR #203, opened 2021-06-07
Release 4.0.0.pre.3

ChrisBAshton:
In the major release which renamed sentry-raven to sentry-ruby,
the rails config was extracted into its own sentry-rails gem:
https://docs.sentry.io/platforms/ruby/guides/rails/

I had hoped to have downstream apps explicitly
`require "sentry-rails"` only if they need it, however this means
the code is evaluated too late in the process, as GovukError is
already defined at that point. For example, trying to set
`config.rails.report_rescued_exceptions` in email-alert-api on
govuk_app_config v4.0.0.pre.2 leads to this error:

```
NoMethodError: undefined method `rails' for #<GovukError::Configuration:0x000055d424a1a8d0>
```

It seems the only alternative is to include `sentry-rails` in
GovukAppConfig by default. Most of our apps are Rails apps anyway,
so in practice this should DRY up a bit of configuration across
our apps, as well as making it easier to keep sentry-rails in sync
with sentry-ruby. The only downside is in Rack apps which don't
have Rails, which now have an unnecessary extra dependency - not
a terrible downside in the grand scheme of things.

Depended on by https://github.com/alphagov/email-alert-api/pull/1629.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
-# 4.0.0.pre-2
+# 4.0.0.pre.3
+
+- Include [sentry-rails](https://github.com/getsentry/sentry-ruby/tree/master/sentry-rails) by default ([#203](https://github.com/alphagov/govuk_app_config/pull/203)).
+
+# 4.0.0.pre.2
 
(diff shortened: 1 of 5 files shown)


thomasleese (approved):
We've already got Rails specific configuration in this repo, so I don't see a problem with including this as well.

PR #205, opened 2021-06-10
Release 4.0.0.pre.4

ChrisBAshton:
This fixes the Sentry client initialisation - see commits for details.

Tested by bumping the dependency in Whitehall and manually sending an [exception](https://sentry.io/organizations/govuk/issues/2447639993/?project=202259&query=&statsPeriod=14d) and [string](https://sentry.io/organizations/govuk/issues/2447640611/?project=202259&query=&statsPeriod=14d) to Sentry, both of which came through 🎉 


lib/govuk_app_config/govuk_error/configuration.rb
```diff
@@ -3,20 +3,82 @@
...
     def initialize(_sentry_configuration)
       super
-      @sentry_environment = ENV["SENTRY_CURRENT_ENV"]
       @data_sync = GovukDataSync.new(ENV["GOVUK_DATA_SYNC_PERIOD"])
-      self.active_sentry_environments = []
-      self.data_sync_excluded_exceptions = []
+      set_up_defaults
+    end
+
+    def set_up_defaults
```
thomasleese, line 15: It occured to me after I wrote the experimental code that this method could belong in `govuk_error.rb`, so we'd have something like this:

```rb
set_up_defaults(config)
yield config if block_given?
```

I think it's slightly nicer because it keeps the `Configuration` class responsible for just wrapping the Sentry class, but to be honest, I don't really mind either way.

thomasleese (approved):

lib/govuk_app_config/govuk_error/configuration.rb
ChrisBAshton: I think having it in `Configuration` is slightly nicer now, to be honest. We can only make one call to `GovukError.configure`, which under the hood is now responsible for merging in custom `before_send` logic, including datasync-error-ignoring, etc. I think it's quite nice that there's just one file to look at all of the default configuration (including the logic above), so there's a little bit less poking around involved if you want to override something in your `GovukError.configure` call.

PR #206, opened 2021-06-16
Add Speedcurve's LUX to connect-src policy

injms:
Add the domain name that Speedcurve's LUX is using for real user metics - this allows LUX to report on how a page is performing.


richardTowers (approved):
Thanks! Looks good.

Might be worth mentioning in the commit message that this is needed in both the image-src and connect-src bits of the policy because we load the speedcurve payload through an ``, but it's served with a non-image content type (`application/javascript`).

richardTowers (approved):

PR #207, opened 2021-06-16
Release v3.2.0

injms:
Includes:

-  Add Speedcurve's LUX to connect-src policy ([#206](https://github.com/alphagov/govuk_app_config/pull/206))


injms:
This was manually released, so this pull request is not needed.

PR #213, opened 2021-10-06
Update reference to klick2contact.com

ChrisBAshton:
Smokey has started failing on `cucumber -p staging_aws features/finder_frontend.feature:90 # Scenario Outline: Check search results and analytics`

This is the error:

```
Refused to connect to 'https://hmpowebchat.klick2contact.com/v03/providers/HMPO/api/availability.php' because it violates the following Content Security Policy directive: "connect-src 'self' *.publishing.service.gov.uk *.staging.publishing.service.gov.uk www.gov.uk *.dev.gov.uk www.google-analytics.com ssl.google-analytics.com stats.g.doubleclick.net www.googletagmanager.com www.tax.service.gov.uk hmrc-uk.digital.nuance.com gov.klick2contact.com www.signin.service.gov.uk lux.speedcurve.com".
```

The content security policy originally included `gov.klick2contact.com` in 7c1670bc1dd88955fed54147178e2c661afe2ca7 in January 2020.
However, we started referencing `hmpowebchat.klick2contact.com` in 1a5adccfd19435c3bceaeab402224775ecbae638 in March 2020.

I'm not sure why this has only started failing today, nor whether `gov.klick2contact.com` is still in use.
My assumption is that that is a legacy thing and we should only allow connections to `hmpowebchat.klick2contact.com` from now on.
If that assumption is wrong, it will be relatively simple to add `gov.klick2contact.com` back in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 4.0.1
 
 - Update Content Security Policy with new klick2contact.com subdomain ([#213](https://github.com/alphagov/govuk_app_config/pull/213)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Update Content Security Policy with new klick2contact.com subdomain ([#213](https://github.com/alphagov/govuk_app_config/pull/213)).
+
 # 4.0.0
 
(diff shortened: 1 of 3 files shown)


bilbof (approved):
Thanks for fixing, I was able to reproduce this issue on https://www.gov.uk/government/organisations/hm-passport-office/contact/hm-passport-office-webchat.

bilbof:
Also, it looks like we discovered this problem by accident (this page happens to be the 11th result in search results for the query "passport" https://github.com/alphagov/smokey/blob/main/features/finder_frontend.feature#L90). Should we add a feature test that this page is working as expected?

PR #216, opened 2021-11-10
Remove Speedcurve's LUX from `connect-src` policy

injms:
## What

Removes the domain the LUX uses for reporting real user metrics from the `connect-src` content security policy.

Reverts #206.

## Why

The domain was originally added because although an image was being used to report the metrics, the image was being served with a content type of `application/javascript`. This meant the loading of the image was being blocked, which prevented any metrics from being reported.

This rule can now be removed from the `connect-src` set of rules as the content type is now `image/webp`:

Screenshot 2021-11-10 at 11 11 46

and is allowed as the domain is in the `img-src` policy:

https://github.com/alphagov/govuk_app_config/blob/a6db8fce8bfaf588053ebb1fe55863043bcc7721/lib/govuk_app_config/govuk_content_security_policy.rb#L36-L38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Remove Speedcurve's LUX from the connect-src policy ([#216](https://github.com/alphagov/govuk_app_config/pull/216)).
+
 # 4.2.0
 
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
Don't forget to rebase for the changelog

injms:
Rebased against `main` and force pushed.

PR #224, opened 2022-03-04
Fix loading gem due missing PrometheusExporter constant

theseanything:
- Fix issue where GovukPrometheusExporter module prevented the gem to load due to missing constant. It was unable to find the top level module name "PrometheusExporter" and need to be explicity required.

- Lazy load the prometheus_exporter dependency for only apps that use GovukPrometheusExporter. This prevents the library from being required by apps the don't need it.

- Load GovukPrometheusExporter only for Rails apps, as it loads middleware into Rack and explicitly depends on Rails being present.


marcpomfret (approved):
Thanks for sorting this.

PR #225, opened 2022-03-09
Update HMPO webchat address in security policy

Tetrino:
As part of a piece of work updating the webchat address in Government Frontend.

https://trello.com/c/wg6cckR1/770-set-up-a-test-page-for-hmpo-to-check-new-webchat-update


chao-xian (approved):

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -75,8 +75,8 @@ def self.build_policy(policy)
                        "www.tax.service.gov.uk",
                        # Allow JSON call to Nuance - HMRC web chat provider
                        "hmrc-uk.digital.nuance.com",
-                       # Allow JSON call to klick2contact - HMPO web chat provider
-                       "hmpowebchat.klick2contact.com",
+                       # Allow JSON call to Eckoh - HMPO web chat provider
+                       "omni.eckoh.uk",
```
leenagupte, line 79: Hmmm. Perhaps you should keep both entries until the new provider goes live?

Tetrino: @leenagupte The new provider is live already, due to communication slowdown we're actually behind HMPO on this rollout.

leenagupte: Sorry, I meant in our code.

leenagupte (approved):
🎉

kevindew:
As this config is only used in one application, government-frontend, I imagine it'd be less friction to manage this via government-frontend itself rather than across all GOV.UK apps. 

Here's an example of prior art of this: https://github.com/alphagov/govuk_publishing_components/blob/86b361c94d13d1566c3a483674faf04df796b1ce/app/controllers/govuk_publishing_components/application_controller.rb#L13-L17

PR #231, opened 2022-03-28
Add flag to only use prometheus exporter on kubernetes cluster

marcpomfret:
Enable prometheus_exporter on EKS only, extra latency and 500 errors caused on EC2 instances running unicorn, don't expect to see the same on EKS with puma 

([Collections 2712](https://github.com/alphagov/collections/pull/2712))Had to be reverted due to increased latency being seen on the controllers of the collections app running on EC2 instances.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix prometheus_exporter to only be enabled when the GOVUK_PROMETHEUS_EXPORTER env var is set to "true" ([#231](https://github.com/alphagov/govuk_app_config/pull/231)).
+- Add Prometheus monitoring for EKS section to README.md ([#231](https://github.com/alphagov/govuk_app_config/pull/231)).
 - Fix govuk_error being incompatible with Ruby >= 3 ([#233](https://github.com/alphagov/govuk_app_config/pull/233))
 - Require Ruby 2.7 as the minimum supported Ruby version ([#233](https://github.com/alphagov/govuk_app_config/pull/233))
 - Require Sentry 5 and Unicorn 6 major versions ([#237](https://github.com/alphagov/govuk_app_config/pull/237))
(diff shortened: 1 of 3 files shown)


lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -1,6 +1,6 @@
 module GovukPrometheusExporter
   def self.configure
-    unless Rails.env == "test"
+    unless Rails.env == "test" || (ENV["GOVUK_PROMETHEUS_EXPORTER"]).blank?
```
theseanything, line 3: Is there an explicit value that we expect "GOVUK_PROMETHEUS_EXPORTER" to be set to? I'm just wondering, is there any case where GOVUK_PROMETHEUS_EXPORTER is "false" or similar

theseanything: So thinking something like might have more predictable behaviour:
```suggestion
    unless Rails.env == "test" || (ENV["GOVUK_PROMETHEUS_EXPORTER"]) != "true"
```

This will make it explicit that we're expecting `GOVUK_PROMETHEUS_EXPORTER = true` to enable the prometheus exporter. If we check for `.blank?` then for example `GOVUK_PROMETHEUS_EXPORTER = false` would also enable the prometheus exporter.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Enable prometheus_exporter on EKS only, extra latency and 500 errors caused on EC2 instances running unicorn, don't expect to see the same on EKS with puma ([#231](https://github.com/alphagov/govuk_app_config/pull/231)).
```
theseanything, line 3:
```suggestion
- Fix prometheus_exporter to only be enabled when the GOVUK_PROMETHEUS_EXPORTER env var is set to "true" ([#231](https://github.com/alphagov/govuk_app_config/pull/231)).
```

We should just detail what exactly the change is - the specifics about why we've done can be found in the commit message.

theseanything (approved):

PR #233, opened 2022-04-05
Use GitHub actions for CI and support Ruby 3

kevindew:
This moves this project from a Jenkins based CI to Github Actions. My motivation for doing so was that I wanted this to be tested with Ruby 3 and 3.1, which our current CI doesn't support, after I found out that this gem is a blocker for GOV.UK apps using Ruby 3. 

More info in the commits.


ChrisBAshton (approved):
Really clear commit structure - thanks!

PR #234, opened 2022-04-05
Only require sentry-rails if Rails is defined

kevindew:
Without this non Rails app will get a warning each time they initialise.

For example, Search API has:

```
W, [2022-03-10T09:45:26.609711 #11663] WARN -- sentry: ** [Sentry] sentry-rails can't detect Rails.logger. it may be caused by misplacement of the SDK initialization code
```


ChrisBAshton (approved):
Good catch.

PR #235, opened 2022-04-05
Some rubocop TLC

kevindew:
It's good to pin this to a particular version of rubocop-govuk,
otherwise we may have different results locally and in CI.

There doesn't seem to be any rationale for the `--format clang` argument
on rubocop so might as well remove it.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,7 @@ RSpec::Core::RakeTask.new(:spec)
 
 desc "Lint Ruby"
 task :lint do
-  sh "bundle exec rubocop --format clang"
+  sh "bundle exec rubocop"
 end
 
 task default: %i[spec lint]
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -5,7 +5,7 @@ RSpec::Core::RakeTask.new(:spec)
 
 desc "Lint Ruby"
 task :lint do
-  sh "bundle exec rubocop --format clang"
```
ChrisBAshton, line 8: I did always wonder what this was for. Thanks for removing!

ChrisBAshton (approved):

Rakefile
kevindew: Yeah - I'm not sure why it ever started, but it has sure spread through our codebase without anyone really understanding (anymore) if it's of value.

PR #236, opened 2022-04-05
Upgrade Rails dev dependency to Rails 7

kevindew:
It's good to test against the current version of Rails. I don't think we
do enough with Rails testing that it is worth to test this with both
Rails 6 and Rails 7. I could be persuaded otherwise if problems occur.

This required an update for the tests as the base active_support
dependency now needs to be required before requiring any extensions.


kevindew:
I'll move this onto https://github.com/alphagov/govuk_app_config/pull/233 since it requires dropping Ruby 2.6

PR #237, opened 2022-04-05
Loosen and bump dependencies

kevindew:
We had set up the gem dependencies quite restrictive which meant apps were quite far behind in some cases (Sentry gems are up to 5.2). This updates the gems to their current versions with flexibility for the releases in the current major version. From looking at the respective change logs, both major version updates were very minor and shouldn't require us to make any app changes.

As this project is used as a way to provide dependencies to GOV.UK apps I've taken the approach that this gem pushes for apps to use newer versions rather than be very flexible in what it accepts - this is a deliberate decision to help keep GOV.UK apps on modern versions of dependencies. 

I've put examples of this into a couple of apps with e2e tests to check they pass all tests with these new versions:

Whitehall: https://github.com/alphagov/whitehall/tree/govuk_app_config-deps
Government Frontend: https://github.com/alphagov/government-frontend/tree/govuk_app_config-deps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Require Sentry 5 and Unicorn 6 major versions ([#237](https://github.com/alphagov/govuk_app_config/pull/237))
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM 👍

PR #241, opened 2022-05-11
Add warnings to integrate GovukError with sentry-sidekiq

kevindew:
Trello: https://trello.com/c/CXTgjI76/383-look-at-sidekiqjobretryskip-errors-logged-to-sentry

This PR looks to address one of the things that has been problematic since the replacement of sentry-raven, how Sidekiq integrates with Sentry. As part of the introduction [sentry-sidekiq](https://docs.sentry.io/platforms/ruby/guides/sidekiq/) was created as a new gem for integrating with Sentry. In my recent experience of sentry alerts and GOV.UK apps, whether an app alerts to sentry from Sidekiq seems to be rather inconsistent. We're seeing some apps alert Sidekiq::JobRetry::Skip exceptions, others seem to alert nothing - it's a bit confusing but installing sidekiq-sentry seems to settle it down.

I don't think it's a good idea to add sentry-sidekiq as a dependency to this gem however as that would mean that all apps that use it would get Sidekiq and a number of other indirect dependencies, which seems both a bit of a pain and a source of confusion. Therefore I've added a warning to prompt apps that use Sidekiq to install sentry sidekiq.

Example:

```
➜  link-checker-api git:(app-config-local) ✗ rails c
Warning: GovukError is not configured to track Sidekiq errors, install the sidekiq-sentry gem to track them.
Loading development environment (Rails 7.0.2.3)
irb(main):001:0>
```

I've also added "Sidekiq::JobRetry::Skip" to the list of excluded exceptions since this is an internal exception [that sidekiq-sentry ignores](https://github.com/getsentry/sentry-ruby/blob/0fe8d27e4f84fda11475437a1cdd16f016efd904/sentry-sidekiq/lib/sentry/sidekiq/configuration.rb#L7-L12) - however because of our custom configuration we seem to need to add this exception to our list manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Add internal Sidekiq exception "Sidekiq::JobRetry::Skip" to excluded exceptions ([#241](https://github.com/alphagov/govuk_app_config/pull/241)).
+
 # 4.5.0
 
 - Add lux.speedcurve.com to connect_src for GOV.UK Content Security Policy ([#232](https://github.com/alphagov/govuk_app_config/pull/232))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 4.6.0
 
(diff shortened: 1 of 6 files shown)


ChrisBAshton (approved):
Thanks for breaking this down commit by commit 👍 LGTM. ⭐

PR #248, opened 2022-06-17
Add new GA domain to security policy

JonathanHallam:
A new domain has caused our [smoke tests to fail](https://deploy.integration.publishing.service.gov.uk/job/Smokey/40512/console) This adds the new domain in which will hopefully fix the issue


1pretz1 (approved):
LGTM

PR #250, opened 2022-06-20
Add region1.google-analytics.com to CSP

gclssvglx:
## What

We should update the [Content Security Policy (CSP)](https://github.com/alphagov/govuk_app_config/blob/4c7198b26acdb096be318c7aa343d46f02b7728d/lib/govuk_app_config/govuk_content_security_policy.rb) with `region1.google-analytics.com`.

## Why

The current CSP is throwing intermittent errors [on integration](https://www.integration.publishing.service.gov.uk/).

## Screenshots

| Before | After |
|---|---|
|Screenshot 2022-06-20 at 08 45 36| |

[Trello](https://trello.com/c/o7LqB0tL/105-update-content-security-policy-csp)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 4.6.3
+
+- Adds `region1.google-analytics.com` to the security policy for GA ([#250](https://github.com/alphagov/govuk_app_config/pull/250))
+
 # 4.6.2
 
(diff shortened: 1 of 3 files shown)


JonathanHallam (approved):
If this works we probably want to remove the www. line. Happy to approve for now to get smokey running properly

PR #253, opened 2022-07-25
WIP - allow CSP to be disabled entirely

richardTowers:
At the moment we only have two settings for our CSP - fully on, or "report-only".

We don't actually have a working report-uri to point our CSP reports at.
It would be better to have no CSP at all in production than to have a broken report-only CSP.

It may be better to have an enum flag like:

GOVUK_CSP_MODE = {"on","report-only","disabled"}

... but that would be a little more work to implement.

This will require an additional change in GOV.UK puppet.


kevindew:
I'm going to optimistically close this since we have a working CSP report URI again.

PR #255, opened 2022-08-03
Add Sidekiq Prometheus instrumentation

fredericfran-gds:
Add Sidekiq Prometheus instrumentation
If Sidekiq is present, add the Sidekiq Prometheus instrumentation.

Ref:
1. [trello card](https://trello.com/c/K8aUtwz4/974-graphs-for-sidekiq-queues-in-grafana)

### DIFF
--- a/lib/govuk_app_config/govuk_prometheus_exporter.rb
+++ b/lib/govuk_app_config/govuk_prometheus_exporter.rb
@@ -1,13 +1,35 @@
 module GovukPrometheusExporter
+  def self.should_configure
+    ENV["GOVUK_PROMETHEUS_EXPORTER"] == "true" && !(defined?(Rails) && Rails.env == "test")
+  end
+
   def self.configure
-    unless Rails.env == "test" || (ENV["GOVUK_PROMETHEUS_EXPORTER"]) != "true"
-      require "prometheus_exporter"
-      require "prometheus_exporter/server"
-      require "prometheus_exporter/middleware"
+    return unless should_configure
 
-      server = PrometheusExporter::Server::WebServer.new bind: "0.0.0.0", port: 9394
-      server.start
+    require "prometheus_exporter"
+    require "prometheus_exporter/server"
+    require "prometheus_exporter/middleware"
 
+    if defined?(Sidekiq)
+      Sidekiq.configure_server do |config|
+        require "prometheus_exporter/instrumentation"
+        config.server_middleware do |chain|
+          chain.add PrometheusExporter::Instrumentation::Sidekiq
+        end
+        config.death_handlers << PrometheusExporter::Instrumentation::Sidekiq.death_handler
+        config.on :startup do
+          PrometheusExporter::Instrumentation::Process.start(type: "sidekiq")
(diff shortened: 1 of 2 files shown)


lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -1,14 +1,34 @@
 module GovukPrometheusExporter
   def self.configure
-    unless Rails.env == "test" || (ENV["GOVUK_PROMETHEUS_EXPORTER"]) != "true"
+    if (ENV["GOVUK_PROMETHEUS_EXPORTER"]) == "true"
+      return if defined?(Rails) ? Rails.env == "test" : false
```
sengi, line 4: Highly recommend factoring the condition out into a separate function to make the logic clearer. Something like

```ruby
def self.should_configure
  ENV['GOVUK_PROMETHEUS_EXPORTER'] == 'true' && defined? Rails && Rails.env != 'test'
end

def self.configure
  return unless should_configure

  require 'prometheus_exporter'
...
```

lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -1,14 +1,34 @@
...
+          end
+        end
+      end
+
       server = PrometheusExporter::Server::WebServer.new bind: "0.0.0.0", port: 9394
       server.start
 
-      Rails.application.middleware.unshift PrometheusExporter::Middleware
+      if defined?(Rails)
+        Rails.application.middleware.unshift PrometheusExporter::Middleware
```
sengi, line 30: Is it preferable to use the [`config.middleware` interface](https://guides.rubyonrails.org/configuring.html#adding-custom-middleware) here rather than manipulating `application.middleware` directly?
```suggestion
        Rails.application.config.middleware.insert_before 0, PrometheusExporter::Middleware
```

lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -1,14 +1,34 @@
...
 
+      if defined?(Sidekiq)
+        Sidekiq.configure_server do |config|
+          require "prometheus_exporter/instrumentation"
+          config.server_middleware do |chain|
+            chain.add PrometheusExporter::Instrumentation::Sidekiq
+          end
+          config.death_handlers << PrometheusExporter::Instrumentation::Sidekiq.death_handler
+          config.on :startup do
+            PrometheusExporter::Instrumentation::Process.start type: "sidekiq"
```
sengi, line 18:
```suggestion
            PrometheusExporter::Instrumentation::Process.start(type: "sidekiq")
```
https://rubystyle.guide/#method-call-parentheses
(similarly throughout)

lib/govuk_app_config/govuk_prometheus_exporter.rb
fredericfran-gds: I copied it across from the documentatio: https://github.com/discourse/prometheus_exporter/blob/main/README.md#rails-integration

kentsanggds (approved):
LGTM

PR #256, opened 2022-08-12
Release version 4.7.0 of the Gem

fredericfran-gds:
This will include #255 where Prometheus Sidekiq monitoring is
added.


fredericfran-gds:
There are other changes but they seem to fall under the categories:
1. update of dependent gems which I don't see people adding in the change log
2. updates some infrastructure code unrelated to the Gem itself

sengi (approved):
👍

PR #260, opened 2022-08-19
Enable EKS environment names for Sentry

theseanything:
This adds environment names for EKS versions of integration, staging and production.


sengi (approved):
LGTM — I think we'll need these in order to tell easily between errors that are unique to one or other hosting environment vs common to both.

sengi:
I think you might need to add these [here](https://github.com/alphagov/govuk_app_config/blob/d61d40c/lib/govuk_app_config/govuk_error/configuration.rb#L20) as well.

PR #261, opened 2022-09-01
Add GovukProxy::StaticProxy for development

theseanything:
This class can be loaded as Rack middleware and will proxy any requests with the path prefix of `/assets/static` to backend specified. This is useful as it enables Static to use relative paths and the dependent  frontend application can forward those requests. This allows us to switch Static to use relative path and allow us to have development environment and not run Static locally (i.e. forward request to production version of Static).

Rails can be made to initialise the proxy middleware by setting the `GOVUK_PROXY_STATIC_ENABLED=true` env var.

### DIFF
--- a/govuk_app_config.gemspec
+++ b/govuk_app_config.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = %w[lib]
 
   spec.add_dependency "logstasher", "~> 2.1"
+  spec.add_dependency "plek", "~> 4"
   spec.add_dependency "prometheus_exporter", "~> 2.0"
   spec.add_dependency "puma", "~> 5.6"
   spec.add_dependency "rack-proxy", "~> 0.7"
--- a/govuk_app_config.gemspec
+++ b/govuk_app_config.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "logstasher", "~> 2.1"
   spec.add_dependency "prometheus_exporter", "~> 2.0"
   spec.add_dependency "puma", "~> 5.6"
+  spec.add_dependency "rack-proxy", "~> 0.7"
   spec.add_dependency "sentry-rails", "~> 5.3"
   spec.add_dependency "sentry-ruby", "~> 5.3"
   spec.add_dependency "statsd-ruby", "~> 1.5"
--- a/lib/govuk_app_config/govuk_proxy/static_proxy.rb
+++ b/lib/govuk_app_config/govuk_proxy/static_proxy.rb
@@ -1,0 +1,19 @@
+require "rack-proxy"
+
+module GovukProxy
+  class StaticProxy < Rack::Proxy
+    def perform_request(env)
+      request = Rack::Request.new(env)
+
+      # use rack proxy to forward any requests for /assets/static/*
+      # this regex needs to match the path set for `Rails.application.config.assets.prefix` in Static
+      # https://github.com/alphagov/static/blob/main/config/initializers/assets.rb
+      if request.path =~ %r{^/assets/static/}
+        env["HTTP_HOST"] = @backend.host
+        super(env)
+      else
+        @app.call(env)
+      end
+    end
+  end
+end
--- a/lib/govuk_app_config/railtie.rb
+++ b/lib/govuk_app_config/railtie.rb
@@ -1,5 +1,14 @@
+require "plek"
+
 module GovukAppConfig
   class Railtie < Rails::Railtie
+    initializer "govuk_app_config.configure_govuk_proxy" do |app|
+      if ENV["GOVUK_PROXY_STATIC_ENABLED"] == "true"
+        static_url = Plek.new.find("static")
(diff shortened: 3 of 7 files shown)


spec/lib/govuk_proxy/static_proxy_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "ignores requests with no path" do
+    test_request("/", false)
+  end
+end
+
+def test_request(path, proxied)
+  dest_host = (proxied ? static_domain : app_domain)
+
+  stub_request(:get, "https://#{dest_host}#{path}")
+      .with(headers: { "Host" => dest_host, "X-Forwarded-For" => "" })
```
sengi, line 34: Would it also be useful to simulate `X-Forwarded-Host` here? (I don't know if it matters, just something to consider)

sengi (approved):

spec/lib/govuk_proxy/static_proxy_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  it "ignores requests where /asset/static isn't a prefix" do
+    test_request("/another/prefix/assets/static/a.css", false)
+  end
+
+  it "ignores requests with no path" do
+    test_request("/", false)
+  end
+end
+
+def test_request(path, proxied)
```
kevindew, line 30: You should define this helper method within the RSpec describe method, as it'd be only scoped to tests there, whereas here it is a global function available to the whole app (defined as a method on `Object`)

spec/lib/govuk_proxy/static_proxy_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+require "govuk_app_config/govuk_proxy/static_proxy"
+require "rack/mock"
+
+RSpec.describe GovukProxy::StaticProxy do
+  # dummy app to validate success
+  let(:app) { ->(_env) { [200, {}, "success"] } }
+  let(:app_domain) { "app.domain" }
+  let(:static_domain) { "static.domain" }
+
+  subject { GovukProxy::StaticProxy.new(app, backend: "https://static.domain", streaming: false) }
```
kevindew, line 11: I think this use of [unnamed subject](https://rspec.rubystyle.guide/#use-subject) would fall foul of RSpec linting rules. I've had a look and they're not on for this app so predictably running them is a bit grim. May want to just change this to 

`let(:proxy) { described_class.new(app, backend: "https://static.domain", streaming: false) }` to not leave linting issues.

kevindew (commented):
Really great idea here Sean - implementation looks nice and simple too which is great 👍 

Only substantial comment is regarding usage of Plek.

spec/lib/govuk_proxy/static_proxy_spec.rb
theseanything: Updated - thanks!

lib/govuk_app_config/govuk_proxy/static_proxy.rb
```diff
@@ -0,0 +1,17 @@
+require "rack-proxy"
+
+module GovukProxy
+  class StaticProxy < Rack::Proxy
+    def perform_request(env)
+      request = Rack::Request.new(env)
+
+      # use rack proxy to forward any requests for /assets/static/*
+      if request.path =~ %r{^/assets/static/}
```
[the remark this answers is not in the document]
theseanything: Comment added

spec/lib/govuk_proxy/static_proxy_spec.rb
theseanything: moved

spec/lib/govuk_proxy/static_proxy_spec.rb
theseanything: So don't really need X-Forwarded-Host here - we aren't doing anything with it the proxy, nor do we need it for any of the request we forward onto Static for development. 

Accidentally left in copied from a template. 🤦🏻

lib/govuk_app_config/railtie.rb
```diff
@@ -1,5 +1,10 @@
 module GovukAppConfig
   class Railtie < Rails::Railtie
+    initializer "govuk_app_config.configure_govuk_proxy" do |app|
+      static_url = ENV["GOVUK_PROXY_STATIC_URL"]
+      app.middleware.use GovukProxy::StaticProxy, backend: static_url if static_url.present?
```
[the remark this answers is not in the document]
theseanything: Yup agree - was a bit clunky having to set to URLs. Switched by to using Plek.

govuk_app_config.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "logstasher", "~> 2.1"
   spec.add_dependency "prometheus_exporter", "~> 2.0"
   spec.add_dependency "puma", "~> 5.6"
+  spec.add_dependency "rack-proxy", "~> 0.7"
```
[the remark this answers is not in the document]
sengi: Ah ignore me — thanks for clearing that up for me!

kevindew (approved):

PR #263, opened 2022-10-21
Avoid INFO logspew from the Sentry gem.

sengi:
The Sentry gem spams stdout with INFO logs like this by default, which is particularly annoying when troubleshooting. We don't need info-level logs from the Sentry integration itself.

    [Transport] Sending envelope with items [sessions]  to Sentry

This change doesn't affect the logging of events to Sentry, only the logging of messages from the Sentry gem itself. We're therefore only interested in log messages which might indicate that something is wrong, not log messages that simply indicate that something expected is happening - hence WARN is the appropriate level here.

See [Sentry Ruby config docs](https://docs.sentry.io/platforms/ruby/configuration/options/) under `logger`.

### DIFF
--- a/lib/govuk_app_config/govuk_error.rb
+++ b/lib/govuk_app_config/govuk_error.rb
@@ -40,6 +40,8 @@ module GovukError
 
     Sentry.init do |sentry_config|
       config = Configuration.new(sentry_config)
+      # Avoid "Sending envelope with items ... to Sentry" logspew on stdout.
+      config.logger.level = Sentry::Logger::WARN
       yield config if block_given?
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 4.10.0
+
+- Reduce log level for the Sentry gem from `INFO` to `WARN` to avoid polluting logs with uninformative messages. This only affects log messages from the Sentry gem itself, which go to `stdout`.
+
 # 4.9.0
 
 - Add GovukProxy::StaticProxy to forward Static asset requests by setting `GOVUK_PROXY_STATIC_ENABLED=true`.([#261](https://github.com/alphagov/govuk_app_config/pull/261))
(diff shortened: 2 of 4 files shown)


theseanything (approved):

theseanything:
We typically add a note to the changelog under the title "Unreleased" - helps with keeping it upto date and makes it easier if someone else does the release.

https://github.com/alphagov/govuk_app_config/pull/261/commits/ef546c39ebccd985227be8e5f43403ef8b826450

Also feel free to add a version bump commit - if you wanna release it immediately

sengi:
Whoops, sorry forgot about the release side of things. I've added a commit to bump the version and add to the changelog. RFAL.

sengi:
Oh dear, lemme fix that test 😂

sengi:
Ok, actually RFAL this time 😅

ChrisBAshton (approved):
Good catch 👍

lib/govuk_app_config/govuk_error.rb
```diff
@@ -40,6 +40,8 @@ def self.configure
 
     Sentry.init do |sentry_config|
       config = Configuration.new(sentry_config)
+      # Avoid "Sending envelope with items ... to Sentry" logspew on stdout.
+      config.logger.level = Sentry::Logger::WARN
```
kevindew, line 44: Do you know if this is reconfiguring the logger for a whole application here? I'm wondering if config.logger is the Rails logger for the whole app.

kevindew: Looks like it might be I'm afraid: 

```
[38, 47] in /Users/kevin.dew/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/govuk_app_config-4.9.0/lib/govuk_app_config/govuk_error.rb
   38:       warn "Warning: GovukError is not configured to track Sidekiq errors, install the sentry-sidekiq gem to track them."
   39:     end
   40:
   41:     Sentry.init do |sentry_config|
   42:       config = Configuration.new(sentry_config)
   43:       require "byebug"; byebug # DEBUG @kevindew
=> 44:       yield config if block_given?
   45:     end
   46:   end
   47: end
(byebug) Rails.logger.object_id == config.logger.object_id
true
```

sengi: Damn. Good catch, thanks! I shouldn't have trusted the docs 🙈 Stand by for bugfix release :)

PR #264, opened 2022-10-21
Fix an ownership bug when configuring the Sentry logger.

sengi:
Sentry's default logger turns out to be a reference to `Rails.logger` by default, despite the [sentry-rails docs] saying that it defaults to an instance of `Sentry::Logger`. So I was unwittingly mutating `Rails.logger` through a shared reference. Oops.

Kudos to super sharp-eyed @kevindew for catching this!

Instead of trying to set the log level for the Sentry gem, just disable the "auto session tracking feature" that generates the spammy log messages, since it turns out we don't use that feature. 

Fixes a bug introduced in #263.

[sentry-rails docs]: https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/

### DIFF
--- a/lib/govuk_app_config/govuk_error.rb
+++ b/lib/govuk_app_config/govuk_error.rb
@@ -40,8 +40,6 @@ module GovukError
 
     Sentry.init do |sentry_config|
       config = Configuration.new(sentry_config)
-      # Avoid "Sending envelope with items ... to Sentry" logspew on stdout.
-      config.logger.level = Sentry::Logger::WARN
       yield config if block_given?
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 4.10.1
+
+- Fix an object ownership/sharing bug where the Rails log level was erroneously being set to `WARN` when initialising Sentry.
+
 # 4.10.0
 
 - Reduce log level for the Sentry gem from `INFO` to `WARN` to avoid polluting logs with uninformative messages. This only affects log messages from the Sentry gem itself, which go to `stdout`.
--- a/lib/govuk_app_config/govuk_error/configuration.rb
+++ b/lib/govuk_app_config/govuk_error/configuration.rb
@@ -74,6 +74,10 @@ module GovukError
         "GdsApi::ContentStore::ItemNotFound",
       ]
 
+      # Avoid "Sending envelope with items ... to Sentry" logspew, since we
+      # don't use Sentry's automatic session tracking.
+      self.auto_session_tracking = false
+
       @before_send_callbacks = [
         ignore_excluded_exceptions_in_data_sync,
         increment_govuk_statsd_counters,
--- a/lib/govuk_app_config/version.rb
+++ b/lib/govuk_app_config/version.rb
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "4.10.0".freeze
+  VERSION = "4.10.1".freeze
 end


lib/govuk_app_config/govuk_error.rb
```diff
@@ -41,6 +41,7 @@ def self.configure
     Sentry.init do |sentry_config|
       config = Configuration.new(sentry_config)
       # Avoid "Sending envelope with items ... to Sentry" logspew on stdout.
+      config.logger = Logger.new($stdout)
```
kevindew, line 44: I don't think we want to do this as we'll lose things like structured logging or the directing to log files in development or test. I think it's right that this logger is the application one.

I'm just going to have a ponder about your problem though

sengi: Hm that's a fair point. How sure are you that these logs are honouring those things right now though? 😅 The unwanted logs that I was seeing were unstructured and going to stdout.

Ruby's `Logger` doesn't seem to have a copy constructor or whatever that's called in the Ruby world. What's the usual way to control a gem's log level?

kevindew: Hmm I've had a bit of a look around. Looks like others have found the message annoying in: https://github.com/getsentry/sentry-ruby/issues/1917

I'm wondering what are sessions? and is it even something we want? I guess they're an indication of data being sent to Sentry. I'm not sure why we're sending data when not erroring 🤔 

I feel like a way we could do your change without things behaving too weirdly is if we did something like

```
config.logger = config.logger.dup
config.logger = Sentry::Logger::WARN
```

It'd still be a bit icky as we'd have a rogue extra logger in the app that wouldn't behave right if you change the global logger.

As an aside I think we may have this code in the wrong place, as this is configuration it should be done in https://github.com/alphagov/govuk_app_config/blob/7a4f323d9728cccbdd014e6209a48ab30db09753/lib/govuk_app_config/govuk_error/configuration.rb#L15 with the other configuration for Sentry.

kevindew: Looking at this: https://docs.sentry.io/platforms/ruby/configuration/releases/#sessions I kinda think we can just turn session tracking off. I don't think we even know we have it let alone use it. 

I know it's a bit of a weird way to solve this problem but I think it's a part of a bunch of extra monitoring that Sentry offers which we don't make any use of.

kevindew: Sorry @sengi I totally hadn't noticed your reply in amongst my spiel.

Yeah `.dup` would probably be the thing you want I reference it a bit earlier.

sengi: Heh, `dup` on an object which [doesn't provide its own implementation of `dup`](https://github.com/ruby/logger/search?q=dup) and manages non-[POD](https://en.wikipedia.org/wiki/Passive_data_structure) resources like files sounds scary to me 😅

If we don't use Sentry's session tracking then I'm happy to go with just turning that off like you suggest 👍

sengi: No worries, GitHub seems to be a bit erratic at updating sometimes 😅

I've removed the ill-fated attempt at controlling the log level and turned off the auto sessions feature as you suggested (in the right place hopefully this time - thanks for that!)

kevindew (approved):
Great, these changes look good to me. 👍 

I've gave it a spin on integration and it seems to sufficiently stop those log writes.

They're guidance is a bit confusing in their docs compared to the code as they seem to actively complain if they can't take the Rails logger :-/ https://github.com/getsentry/sentry-ruby/blob/d3d35689a4529942b190001936bd14848b7a6ba7/sentry-rails/lib/sentry/rails/configuration.rb#L14-L23

sengi:
Thanks!

PR #274, opened 2022-12-28
Spring clean of the Content Security Policy

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

This updates the Content Security Policy in a number of ways:

- It removes legacy declarations: data fonts, various webchat hosts and a verify script
- It simplifies the base CSP rule to be only the current host, whereas previously it was any https script, which seemed rather broad.
- It drops setting a variety of asset hosts, this is no longer needed in production as load all assets with relative paths
- It updates documentation on usage and current intentions.
- It adds https://img.youtube.com as an image-src to support the [YouTube card component](https://github.com/alphagov/govuk_publishing_components/pull/3156)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* CSP only allows scripts, styles and fonts from self which reflects GOV.UK production behaviour
 * Set the default CSP behaviour to be allow communication only to self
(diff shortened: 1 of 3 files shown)


chao-xian (approved):
All makes sense to me

PR #278, opened 2023-01-12
Fix inconsistent bullets in changelog

kevindew:
We were all in on asterisks until version 3.1 and then went with dashes.


ChrisBAshton (approved):
Apologies for breaking the trend 😅

PR #279, opened 2023-01-13
Configure Content Security Policy to remove risky allowed properties

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

This applies a number of changes to the Content Security Policy provided by this gem. It:

- removes unsafe_inline from script-src ~and style-src~, which means that any usages of `<script>` ~or `<style>`~ elements requires a nonce ~and usage of inline styles are forbidden~. This is to limit the damage an attacker can do with ability to inject HTML into a page. (Note: I wasn't able to disable unsafe-inline on style due to [Govspeak](https://github.com/alphagov/govspeak/blob/5642fcc4231f215d1c58ad7feb30ca42fb8cfb91/lib/govspeak/html_sanitizer.rb#L72-L73) usage) 
- removes ability to embed images with `data:` URLs. This limits any injected HTML to not display arbitrary images
- disallows a `<base>` element. This limits the ability of injected HTML to change the destination of scripts and allow execution of 3rd party JavaScript

This has been opened as a draft while frontend applications are checked for blocking issues:

- [x] collections (uses inline script tags for structured data, nonces not required for `<script type="application/ld+json">`)
- [x] email-alert-frontend
- [x] feedback
- [x] finder-frontend (fixed in https://github.com/alphagov/finder-frontend/pull/2976)
- [x] frontend (uses inline script tags for structured data, nonces not required for `<script type="application/ld+json">`)
- [x] government-frontend (fixed in https://github.com/alphagov/government-frontend/pull/2661)
- [x] info-frontend
- [x] licence-finder (fixed in https://github.com/alphagov/licence-finder/pull/1280)
- [x] service-manual-frontend
- [x] smart-answers (fixed in https://github.com/alphagov/smart-answers/pull/6277)
- [x] static (has some inline script tags in error pages (not under CSP) and old files that don't seem to be used on public GOV.UK)
- [x] whitehall (does not use CSP currently)

This is intended to be the draft of the CSP that GOV.UK applies as its initial live offering. Hopefully we will be ready to go once the above list is completed and that we've run this on apps in report only mode for a couple of weeks.

There are some future things I considered out of scope for now:

- [`strict-dynamic`](https://content-security-policy.com/strict-dynamic/) - we probably should switch to `strict-dynamic` for scripts in time, as it is a clear improvement. However, as you have to set-up a bunch of extra backwards compatibility changes, I think it makes things extra complicated when we may have some apps still needing things like unsafe-inline for legacy jQuery. Therefore I think this is something to dig into once live.
- nonces in all script elements, we might want to reduce our script-src allow list and just set nonce attributes, indeed https://csp-evaluator.withgoogle.com/ raises concerns that some of our allow-list could be hosts at risk of serving dubious scripts. This wouldn't be achiveable without adopting `strict-dynamic` so we'll reserve this for that.
- [`require-trusted-types-for 'script'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) this is recommended by https://csp-evaluator.withgoogle.com/. Implementation looks very difficult, as it requires identifying every case of `innerHTML` being called, likely extra challenging because of incomplete browser support currently. So I don't see us being able to support this in the forseeable future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* BREAKING: Content Security Policy forbids unsafe-inline script-src and data: image-src. It provides a nonce generator. Apps that can't support this will need to amend their CSP configuration in an initializer, see [example](https://github.com/alphagov/signon/commit/ddcf31f5c30b8fd334e4aea74986b24bf2b0e9be) in signon. Any apps that still use jQuery 1.x will need unsafe-inline for Firefox compatibility.
+
 # 4.13.0
 
 * Flush log writes to stdout immediately so that structured (JSON) logs are not lost on crash or delayed indefinitely.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+* Forbid base elements in the Content Security Policy
 * BREAKING: Content Security Policy forbids unsafe-inline script-src and data: image-src. It provides a nonce generator. Apps that can't support this will need to amend their CSP configuration in an initializer, see [example](https://github.com/alphagov/signon/commit/ddcf31f5c30b8fd334e4aea74986b24bf2b0e9be) in signon. Any apps that still use jQuery 1.x will need unsafe-inline for Firefox compatibility.
 
 # 4.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 5.0.0
 
 * Forbid base elements in the Content Security Policy
 * BREAKING: Content Security Policy forbids unsafe-inline script-src and data: image-src. It provides a nonce generator. Apps that can't support this will need to amend their CSP configuration in an initializer, see [example](https://github.com/alphagov/signon/commit/ddcf31f5c30b8fd334e4aea74986b24bf2b0e9be) in signon. Any apps that still use jQuery 1.x will need unsafe-inline for Firefox compatibility.
--- a/lib/govuk_app_config/govuk_content_security_policy.rb
+++ b/lib/govuk_app_config/govuk_content_security_policy.rb
@@ -29,12 +29,8 @@ module GovukContentSecurityPolicy
     policy.default_src :self
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src
+    # Note: we purposely don't include `data:` here because it produces a security risk.
     policy.img_src :self,
-                   # This allows Base64 encoded images, but is a security
-                   # risk as it can embed third party resources.
-                   # As of December 2022, we intend to remove this prior
-                   # to making the CSP live.
-                   :data,
                    *GOVUK_DOMAINS,
                    *GOOGLE_ANALYTICS_DOMAINS, # Tracking pixels
                    # Speedcurve real user monitoring (RUM) - as per: https://support.speedcurve.com/docs/add-rum-to-your-csp
@@ -45,25 +41,28 @@ module GovukContentSecurityPolicy
                    "https://img.youtube.com"
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
+    # Note: we purposely don't include `data:`, `unsafe-inline` or `unsafe-eval` because
+    # they are security risks, if you need them for a legacy app please only apply them at
+    # an app level.
     policy.script_src :self,
                       *GOOGLE_ANALYTICS_DOMAINS,
                       *GOOGLE_STATIC_DOMAINS,
                       # Allow YouTube Embeds (Govspeak turns YouTube links into embeds)
                       "*.ytimg.com",
                       "www.youtube.com",
-                      "www.youtube-nocookie.com",
-                      # This allows inline scripts and thus is a XSS risk.
-                      # As of December 2022, we intend to work towards removing
-                      # this from apps that don't use jQuery 1.12 (which needs
-                      # this) once we've set up nonces.
-                      :unsafe_inline
+                      "www.youtube-nocookie.com"
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src
+    # Note: we purposely don't include `data:` or `unsafe-eval` because
+    # they are security risks, if you need them for a legacy app please only apply them at
+    # an app level.
     policy.style_src :self,
                      *GOOGLE_STATIC_DOMAINS,
-                     # This allows style="" attributes and style elements.
-                     # As of December 2022, we intend to remove this prior
-                     # to making the CSP live due to the security risks it has.
+                     # This allows `style=""` attributes and `<style>` elements.
+                     # As of January 2023 our intentions to remove this were scuppered
+                     # by Govspeak [1] using inline styles on tables. Until that
+                     # is resolved we'll keep unsafe_inline
+                     # [1]: https://github.com/alphagov/govspeak/blob/5642fcc4231f215d1c58ad7feb30ca42fb8cfb91/lib/govspeak/html_sanitizer.rb#L72-L73
                      :unsafe_inline
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/font-src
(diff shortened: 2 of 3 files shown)


kevindew:
I've amended this to leave unsafe-inline for style. This was done after finding out that is output by [Govspeak](https://github.com/alphagov/govspeak/blob/5642fcc4231f215d1c58ad7feb30ca42fb8cfb91/lib/govspeak/html_sanitizer.rb#L72-L73), [via Kramdown](https://kramdown.gettalong.org/syntax.html#tables), for text alignment in tables.

I have an idea on how to fix this (convert this to a class in Govspeak post-processing) but I still need to write that and then do a rather tedious republishing process to apply it to existing places.

kevindew:
I've tested this on integration with all the frontend apps (those listed above) deployed with it. All Smokey tests continue to pass, I've checked individual pages on each app and no warnings.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+* BREAKING: Content Security Policy forbids unsafe-inline script-src and data: image-src. It provides a nonce generator. Apps that can't support this will need to amend their CSP configuration in an initializer.
```
ChrisBAshton, line 4: Is it worth linking to a doc explaining more about this initializer?

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -45,25 +41,28 @@ def self.build_policy(policy)
                    "https://img.youtube.com"
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
+    # Note: we purposely don't include data:, unsfafe-inline or unsafe-eval because
```
ChrisBAshton, line 47:
```suggestion
    # Note: we purposely don't include data:, unsafe-inline or unsafe-eval because
```

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -45,25 +41,28 @@ def self.build_policy(policy)
...
     policy.script_src :self,
                       *GOOGLE_ANALYTICS_DOMAINS,
                       *GOOGLE_STATIC_DOMAINS,
                       # Allow YouTube Embeds (Govspeak turns YouTube links into embeds)
                       "*.ytimg.com",
                       "www.youtube.com",
-                      "www.youtube-nocookie.com",
-                      # This allows inline scripts and thus is a XSS risk.
-                      # As of December 2022, we intend to work towards removing
-                      # this from apps that don't use jQuery 1.12 (which needs
```
ChrisBAshton, line 57: Also is it worth keeping this reference somewhere (e.g. the change note)? Might be a useful prompt for devs to check what version of jQuery they're running...?
According to the dependency management spreadsheet, most jQuery has been removed, but [not everywhere](https://trello.com/c/7faUW2yP/177-govuk-depends-on-an-eol-version-of-jquery-112).

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -91,6 +90,16 @@ def self.build_policy(policy)
   def self.configure
     Rails.application.config.content_security_policy_report_only = ENV.include?("GOVUK_CSP_REPORT_ONLY")
 
+    # Sets a nonce per request that is set in the CSP directives for script_src and
+    # style_src.
+    # Note: if an applications needs to set unsafe-inline they will need to unset this nonce too (by setting
```
ChrisBAshton, line 98:
```suggestion
    # Note: if an application needs to set unsafe-inline they will need to unset this nonce too (by setting
```

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -91,6 +90,16 @@ def self.build_policy(policy)
...
+    # this config option to nil)
+    Rails.application.config.content_security_policy_nonce_generator = ->(_request) { SecureRandom.base64(16) }
+
+    # This only applies the nonce generator to the script-src directive. We need this to
+    # use unsafe-inline for style-src as a nonce will override it.
+    Rails.application.config.content_security_policy_nonce_directives = %w[script-src]
```
ChrisBAshton, line 104: In other words, once we can apply unsafe-inline to the style dir, we can remove this config line (as I assume it would apply to both by default)?
Or would we be adding style-src to this array?
Would be good to make the comment clearer about what will happen in future.

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -91,6 +90,16 @@ def self.build_policy(policy)
   def self.configure
     Rails.application.config.content_security_policy_report_only = ENV.include?("GOVUK_CSP_REPORT_ONLY")
 
+    # Sets a nonce per request that is set in the CSP directives for script_src and
+    # style_src.
```
ChrisBAshton, line 97: Is it worth adding a note like "(but it will have no effect on the latter until we remove unsafe-inline support for it)"?

ChrisBAshton (approved):
LGTM, some nitpick comments below.

Very thorough PR description, lots of testing & due diligence done - nice one ⭐

lib/govuk_app_config/govuk_content_security_policy.rb
kevindew: Good shout. I thought this was a non-concern as I ensured the removal of jQuery 1.x on all frontend apps.

I, however, had no idea that that in the 2020 Rails update work a bunch of admin applications had the Content Security Policy switched on (including oddly some API only apps): https://sourcegraph.com/search?q=context:global+repo:alphagov+GovukContentSecurityPolicy&patternType=standard&sm=0&groupBy=repo

CHANGELOG.md
kevindew: Yup, that's a good shout. I thought in practice that no apps would need to do anything since I'd sorted them out the frontend apps. But now I've discovered the leak into the admin apps that seems a good idea.

kevindew: I'm going to get an example together before updating this item.

kevindew: I've applied this now.

lib/govuk_app_config/govuk_content_security_policy.rb
kevindew: Should now be covered in the changelog not, but I've been and swept through apps already to have them ready.

kevindew:
Thanks for the review @ChrisBAshton, some great spots. Should be ready for another look.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+* BREAKING: Content Security Policy forbids unsafe-inline script-src and data: image-src. It provides a nonce generator. Apps that can't support this will need to amend their CSP configuration in an initializer, see [example](https://github.com/alphagov/signon/commit/ddcf31f5c30b8fd334e4aea74986b24bf2b0e9be) in signon. Any apps that still use jQuery 1.x will need unsafe-inline for Firefox compatibility.
```
ChrisBAshton, line 4: Nice 👍

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -29,12 +29,8 @@ def self.build_policy(policy)
     policy.default_src :self
 
     # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src
+    # Note: we purposely don't include data here because it produces a security risk.
```
ChrisBAshton, line 35: Suggestion for clarity:

```suggestion
    # Note: we purposely don't include `data:` here because it produces a security risk.
```

lib/govuk_app_config/govuk_content_security_policy.rb
```diff
@@ -91,6 +90,20 @@ def self.build_policy(policy)
   def self.configure
     Rails.application.config.content_security_policy_report_only = ENV.include?("GOVUK_CSP_REPORT_ONLY")
 
+    # Sets a nonce per request that can be set on script-src and style-src
+    # directives depending on the value of Rails.application.config.content_security_policy_nonce_directives
+    #
+    # Note: if an applications needs to set unsafe-inline they will need to
```
ChrisBAshton, line 99:
```suggestion
    # Note: if an application needs to set unsafe-inline they will need to
```

ChrisBAshton (approved):
Thanks Kevin, much clearer! Couple of nitpicks below, but good to merge 👍

lib/govuk_app_config/govuk_content_security_policy.rb
kevindew: Good spot!

PR #282, opened 2023-02-20
Allow us to force-load the Prometheus Exporter...

dj-maisy:
## What?
The Exporter currently ignores the `GOVUK_PROMETHEUS_EXPORTER` environment variable if it detects it is being called by rake.

As the email-alerts-service is a perpetual rake task, we need a way to override this for the time being.

This should allow us to set the `GOVUK_PROMETHEUS_EXPORTER` to `"force"` to make it configure the Prometheus Exporter regardless of whether it is being called by `rake` or not.

## Related PRs

* https://github.com/alphagov/email-alert-service/pull/617


theseanything (approved):

theseanything (approved):
You need to add a line in the changelog. You can also include a commit to bump the version to trigger a release.

PR #286, opened 2023-03-23
Release v7.0.0: drop unicorn and deprecate statsd.

sengi:
- Remove GovukUnicorn and drop unicorn as a dependency. All our production apps are now using Puma.
- Announce the deprecation of GovukStatsd and support for statsd metrics. Most of our instrumentation is already Prometheus, with only a few statsd counters remaining.
- Some documentation cleanup along the way.


sihugh:
Sorry Chris, I released v6 while you were off 😬

robinjam (approved):

PR #288, opened 2023-04-05
Allow puma timeout to be configurable

BeckaL:
This will allow us to set a different timeout for production systems. The need for this arose because of timing out requests for HMRC manuals api, which we [allowed for configuration of](https://github.com/alphagov/govuk_app_config/blob/c68fb5816906bde162d95e8c2738cde9f8f1d1d1/lib/govuk_app_config/govuk_unicorn.rb#L5) when it was unicorn- rather than puma-based.

I'll release this in a follow-up PR.

[Trello](https://trello.com/c/qtpjC3PB/556-increase-timeout-for-hmrc-manuals-api)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 6.0.1
+
+* Add support for configuring timeouts for puma-based applications
+
 ## 6.0.0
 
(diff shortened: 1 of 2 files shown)


sengi (commented):
Thanks for adding this! 🙇

PR #290, opened 2023-04-19
7.1.0: Remove hardcoded allowlist of Sentry environment names.

sengi:
`active_sentry_environments` was a hard-coded list of allowable values for the Sentry `environment` tag. It was [introduced](https://www.github.com/alphagov/govuk_app_config/pull/168) during the Replatforming discovery phase in 2020 in order to prevent the Puppet/EC2 stack in the test account from sending events to Sentry.

(It was done this way because at that time, disabling Sentry in the test environment was more difficult than it should have been, while some teams had Sentry alerts and Slack subscriptions which didn't select by environment).

We don't need this feature any more and we're now better off without it, because it complicated the semantics of `SENTRY_CURRENT_ENV` in a way that is surprising to the user.

This fixes a current problem where apps in the integration cluster are unable to send events to Sentry.

If we want to run without Sentry in future, we can just not set `SENTRY_DSN`.

Rollout: requires all the affected apps to update to govuk_app_config 7.1 (i.e. this release) in order for the fix to take effect :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,17 @@
+# 7.1.0
+
+* `GovukError` now allows specifying any name for the Sentry environment tag via the `SENTRY_CURRENT_ENV` environment variable. The environment name no longer has to match one of a fixed set of strings in order for `GovukError` to log events to Sentry.
+
 # 7.0.0
 
 * BREAKING: Remove [unicorn](https://rubygems.org/gems/unicorn/) and `GovukUnicorn`. All production GOV.UK apps are now using [Puma](https://rubygems.org/gems/puma/) instead.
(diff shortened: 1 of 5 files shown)


sengi:
Could backport this into a 6.1 release as well if we need to get this out sooner, maybe? (Then again, maybe the Sentry bug will encourage teams to update to v7 😈😆)

sengi:
On second thoughts, if we want to get the fix out asap then it's gonna be easiest just to update everyone to v7 en masse; it's a pretty straightforward upgrade (just removes Unicorn, which nobody's actually using any more afaik)

KludgeKML (approved):
Yep, looks good to me!

PR #291, opened 2023-04-27
Forbid unsafe-inline for style attributes in CSP

kevindew:
This continues the work from https://github.com/alphagov/govuk_app_config/pull/279 to remove risky properties from our Content Security Policy (CSP) by removing unsafe-inline from style properties.

We have been able to resolve the need for this property by updating Govspeak [1].

I've marked this as a breaking change as it forbids something fundamental, however this is really quite a soft breaking change as a) there has been lots of prep work across GOV.UK to prepare for it and b) the CSP is in report only mode across GOV.UK so won't actually block any usages.

~Prior to merging this we need to represent any content that uses tables with text alignment to the Publishing API to remove their inline styles. I've opened this as a draft while this is completed.~
Edit: 2 June, I've been through and removed nearly all usages of style. I'm now going to utilise the CSP reports to catch other ones. 

[1]: https://github.com/alphagov/govspeak/pull/268

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 8.0.0
 
 * BREAKING: Content Security Policy forbids the use of inline style attributes.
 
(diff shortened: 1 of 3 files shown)


ChrisBAshton (approved):
LGTM 👍

PR #292, opened 2023-05-09
Add Puma::HttpParserError to excluded exceptions

BeckaL:
Add HttpParserError to excluded exceptions
    
Sentry ruby 5.9 ensured that low level puma errors were captured and propagated to sentry. This ended up in a lot of noise [1] in sentry when applications were bumped to this version of sentry ruby, particularly with Puma::HttpParser errors.
    
These errors are often unactionable and we believe a result of crawler activity. This suppresses these errors in order to reduce sentry noise.
    
1: https://govuk.sentry.io/issues/4154286086/?project=202217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Suppress noisy Puma::HttpParserError errors
+
 # 7.1.0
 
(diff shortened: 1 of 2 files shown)


sihugh (approved):
These do look like they're being generated as a result of crawlers or scanners and not legitimate user activity. There's no evidence of these errors in our inter-app communication

PR #294, opened 2023-05-10
Allow prometheus binding to fail with a warning rather than a crash

KludgeKML:
- This allows us to (eg) run rake tasks or db migrations in situations where the port is already bound (for instance in a pod shell)

### DIFF
--- a/lib/govuk_app_config/govuk_prometheus_exporter.rb
+++ b/lib/govuk_app_config/govuk_prometheus_exporter.rb
@@ -34,15 +34,19 @@ module GovukPrometheusExporter
       end
     end
 
-    server = PrometheusExporter::Server::WebServer.new bind: "0.0.0.0", port: 9394
-    server.start
+    begin
+      server = PrometheusExporter::Server::WebServer.new bind: "0.0.0.0", port: 9394
+      server.start
 
-    if defined?(Rails)
-      Rails.application.middleware.unshift PrometheusExporter::Middleware
-    end
+      if defined?(Rails)
+        Rails.application.middleware.unshift PrometheusExporter::Middleware
+      end
 
-    if defined?(Sinatra)
-      Sinatra.use PrometheusExporter::Middleware
+      if defined?(Sinatra)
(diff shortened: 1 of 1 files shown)


lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -34,15 +34,19 @@ def self.configure
...
+      if defined?(Rails)
+        Rails.application.middleware.unshift PrometheusExporter::Middleware
+      end
 
-    if defined?(Sinatra)
-      Sinatra.use PrometheusExporter::Middleware
+      if defined?(Sinatra)
+        Sinatra.use PrometheusExporter::Middleware
+      end
+    rescue Errno::EADDRINUSE => e
```
sengi, line 48: Would it be safer to put the rescue just after the call to `server.start` on line 39? I think as-is it'll eat exceptions from the part where it registers itself with the web framework (unlikely to throw anything, but if it does it'll be super confusing to debug!)

sengi (approved):
Yay thanks for fixing this! ❤️

Just one minor comment but have a preemptive approval cos I don't wanna slow you down, especially as you've kindly gone out of your way to fix this for us! 🙇

lib/govuk_app_config/govuk_prometheus_exporter.rb
sengi: Oh wait no it's no big deal cos we're only catching EADDRINUSE, duh. Really just a nit then I guess. I'm fine either way.

KludgeKML: Hmm, good point - It felt better to put the rescue late because if we can't bind we also want to not try to start the server, I assume (nor attach the middleware). Bear in mind that our rescue is only one specific type of error, I'd assume we would never see that specific error in the server.start call.

sengi: Sorry for the noise

PR #297, opened 2023-06-09
Fix dropped logs due to incorrect source field type

theseanything:
The logstash gem adds a field to logs called "source" and by defaults sets it a string representing the host IP address. Our Elasticsearch mapping expect source to be of type object, to support subfield such as source.ip or source.address etc. This causes Elasticsearch to drop the logs resulting in missing Rails logs.

There's currently no way to rename the field (as the field is added after the logstasher's renaming config) and it cannot be disabled. This field is of limited use as the value is usually just the K8s pod IP address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 8.0.1
+
+* Change the "source" field in Rails logs from logstasher from string representing IP host address to an empty object. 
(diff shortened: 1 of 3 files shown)


sengi (approved):
Awesome! Thanks for tracking this down!

Issue #298, opened 2023-06-14
Sidekiq workers fail to export Prometheus metrics


sengi:
In at least some use cases, Sidekiq workers fail to initialise PrometheusExporter. Logs on startup look like:

```
PrometheusExporter::Instrumentation::SidekiqProcess Prometheus Exporter Failed To Collect Stats uninitialized
PrometheusExporter::Instrumentation::SidekiqQueue Prometheus Exporter Failed To Collect Stats uninitialized
PrometheusExporter::Instrumentation::SidekiqStats Prometheus Exporter Failed To Collect Stats uninitialized constant Sidekiq::Stats
```

reported-by: @AgaDufrat

PR #299, opened 2023-06-15
Require sidekiq/api in GovukPrometheusExporter

AgaDufrat:
## Why

The workers were erroring with:
```
PrometheusExporter::Instrumentation::SidekiqProcess Prometheus Exporter Failed
To Collect Stats uninitialized constant Sidekiq::Process
PrometheusExporter::Instrumentation::SidekiqQueue Prometheus Exporter Failed To
Collect Stats uninitialized constant Sidekiq::Queue
PrometheusExporter::Instrumentation::SidekiqStats Prometheus Exporter Failed To
Collect Stats uninitialized constant Sidekiq::Stats
```

and failing to send metrics following a switch from Statsd to Prometheus Exporter: https://github.com/alphagov/govuk_sidekiq/commit/763144dd055ab286a020902e7d63a21eec428155

Sidekiq::Process, Sidekiq::Queue, Sidekiq::Stats are defined in Sidekiq API. It appears that to get access to the functionality, the API needs to be required. See https://github.com/sidekiq/sidekiq/wiki/API

Resolves https://github.com/alphagov/govuk_app_config/issues/298

## Testing
Tested in integration. Logs showing no errors

<kbd>Screenshot 2023-06-15 at 11 19 20</kbd>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 8.0.2
 
 * Fix the ability to collect Sidekiq metrics in GovukPrometheusExporter. ([#299](https://github.com/alphagov/govuk_app_config/pull/299))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Fix the ability to collect Sidekiq metrics in GovukPrometheusExporter. ([#299](https://github.com/alphagov/govuk_app_config/pull/299))
+
 # 8.0.1
 
 * Change the "source" field in Rails logs from logstasher from string representing IP host address to an empty object. 
--- a/lib/govuk_app_config/govuk_prometheus_exporter.rb
+++ b/lib/govuk_app_config/govuk_prometheus_exporter.rb
@@ -20,6 +20,7 @@ module GovukPrometheusExporter
 
     if defined?(Sidekiq)
       Sidekiq.configure_server do |config|
+        require "sidekiq/api"
         require "prometheus_exporter/instrumentation"
         config.server_middleware do |chain|
           chain.add PrometheusExporter::Instrumentation::Sidekiq
--- a/lib/govuk_app_config/version.rb
+++ b/lib/govuk_app_config/version.rb
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "8.0.1".freeze
+  VERSION = "8.0.2".freeze
 end


lib/govuk_app_config/govuk_prometheus_exporter.rb
```diff
@@ -14,6 +14,7 @@ def self.should_configure
   def self.configure
     return unless should_configure
 
+    require "sidekiq/api"
```
sengi, line 17: Heh, this is a slightly tricky one because there's a ton of iffy dynamic loading stuff going on here :)

This library needs to work with apps that don't use Sidekiq as well as ones that do. So if we're trying to fix an undeclared conditional dependency of `prometheus_exporter` on `sidekiq/api` I think we'd need to put the conditional inside the `if defined?(Sidekiq)` below. i.e. just underneath what's currently line 22.

AgaDufrat: I tried that but it didn't work. I think it needs to be required before `require "prometheus_exporter"` 🤔

AgaDufrat: Maybe we should go for `require "sidekiq/api" if defined?(Sidekiq)` on line 17?

sengi: That's ok but slightly less desirable, because you're duplicating the condition so there's just one more thing to go wrong as the file changes over time.

sengi (approved):

lib/govuk_app_config/govuk_prometheus_exporter.rb
AgaDufrat: I just tried it again and it worked in that if block. _(My terminal history confirms I tired it before. Maybe I haven't restarted the worker... anyway.)_ I'll pop it in before `require "prometheus_exporter/instrumentation"` so they are all nicely together. That is much better – thank you!

sengi: No worries!

sengi:
Thanks for fixing this! 🙇✨

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
sengi, line 1: Totes optional: in case you wanna get this out quickly, feel free to do the 8.0.2 release in this PR (entirely up to you though, just mentioning in case it saves you time/hassle of getting another review for essentially the same change)

PR #300, opened 2023-06-22
Log govuk request id in json in whitehall

yuetylauiris:
Change Rails logger to log in JSON format including govuk_request_id for traceability.
Remove the $real_stdout, $stdout and $stderr redirection because LOGIT is able to consume not JSON logs now.

Related Trello card: https://trello.com/c/cfTa0MXn/58-story-log-govukrequestid-from-railslogger-in-whitehall-and-its-worker

PR #302, opened 2023-06-30
Enable production-like logs if env variable is set

lauraghiorghisor-tw:
This is to allow setting production-like behaviour by using the `ENABLE_PRODUCTION_LIKE_LOGS` env variable. This will be set in `govuk-heml-charts` -> `env-configmap.yaml`. The purpose is to allow developers to override this env var in order to test changes to production logs locally.

[Trello card](https://trello.com/c/niLCNflj/117-story-enable-production-like-logging-in-dev-environments)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* BREAKING: JSON logs are no longer configured automatically for production Rails apps and are turned on with the GOVUK_RAILS_JSON_LOGGING environment variable ([#302](https://github.com/alphagov/govuk_app_config/pull/302))
+
 # 8.1.1
 
 * Fix prometheus_exporter to method patching compatible with open telemetry.
--- a/README.md
+++ b/README.md
@@ -145,6 +145,13 @@ check docs](docs/healthchecks.md) for more information on how to use it.
 In Rails applications, the application will be configured to send JSON-formatted
 logs to `STDOUT` and unstructured logs to `STDERR`.
 
+To enable production-like logging, an env variable `GOVUK_RAILS_JSON_LOGGING`
+is set in the `govuk-helm-charts` and then checked in `railtie.rb`. This will
+allow JSON format logs and `Govuk-Request-Id` to be visible.
+
+For development logs, in order to see the production style logs, developers should
+set `GOVUK_RAILS_JSON_LOGGING`in `govuk-docker` -> `docker-compose` files.
+
 
 ## Content Security Policy generation
 
--- a/lib/govuk_app_config/railtie.rb
+++ b/lib/govuk_app_config/railtie.rb
@@ -13,7 +13,7 @@ module GovukAppConfig
     end
 
     config.before_initialize do
-      GovukLogging.configure if Rails.env.production?
+      GovukJsonLogging.configure if ENV["GOVUK_RAILS_JSON_LOGGING"]
     end
 
     config.after_initialize do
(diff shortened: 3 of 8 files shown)


lib/govuk_app_config/railtie.rb
```diff
@@ -9,7 +9,7 @@ class Railtie < Rails::Railtie
     end
 
     config.before_initialize do
-      GovukLogging.configure if Rails.env.production?
+      GovukLogging.configure if ENV["ENABLE_PRODUCTION_LIKE_LOGS"]
```
kevindew, line 12:
```suggestion
      GovukLogging.configure if ENV["GOVUK_RAILS_JSON_LOGGING"]
```

I think we should name this more based on what it does and we can have the naming [be consistent with govuk_sidekiq](https://github.com/alphagov/govuk_sidekiq/blob/26f3ff2cf500f507fc9b34089c811728365df627/lib/govuk_sidekiq/sidekiq_initializer.rb#L22) - which feels a bit more 12 factor.

I imagine a cherry on top would be to rename the class to GovukJsonLogging which would reflect the changes of https://github.com/alphagov/govuk_app_config/pull/300 where GovukLogging is now only setting up JSON logging

kevindew (commented):
I've made a suggestion to the variable naming.

We'll also need to add an entry to the change log. With semantic versioning we should consider this a breaking change, which would seem a bit of a pain but given https://github.com/alphagov/govuk_app_config/pull/300 already has breaking changes we can release them together and not need two major versions.

I'll leave a note on https://github.com/alphagov/govuk_app_config/pull/300 to say to be careful with their version bump.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 9.0.0
+
+* BREAKING: Enable production-like logs if env variable is set ([#302](https://github.com/alphagov/govuk_app_config/pull/302))
```
theseanything, line 3: Not sure I fully understood what you mean by "production-like" logs. Is it switch the logs to JSON formatting etc?

lauraghiorghisor-tw: @theseanything I think Kevin's comment above explains this best. Yes, it switches the logs to JSON - the main chunk of work is actually in #300

theseanything: (Sorry just realised Kevin left a similar comment)

README.md
```diff
@@ -145,6 +145,13 @@ check docs](docs/healthchecks.md) for more information on how to use it.
 In Rails applications, the application will be configured to send JSON-formatted
 logs to `STDOUT` and unstructured logs to `STDERR`.
```
kevindew, line 146: Ooh we should update this in https://github.com/alphagov/govuk_app_config/pull/300 /cc @yuetylauiris

kevindew (approved):
Looks good to me, thanks. I guess we can't release this until we've got the helm chart change applied to each env

floehopper:
> I guess we can't release this until we've got the helm chart change applied to each env

@lauraghiorghisor-tw Can I double-check that https://github.com/alphagov/govuk-helm-charts/pull/1190 adds the `ENABLE_PRODUCTION_LIKE_LOGS` env var to all apps and so it's safe to merge the latest version of this gem?

lauraghiorghisor-tw:
> > I guess we can't release this until we've got the helm chart change applied to each env
> 
> @lauraghiorghisor-tw Can I double-check that [alphagov/govuk-helm-charts#1190](https://github.com/alphagov/govuk-helm-charts/pull/1190) adds the `ENABLE_PRODUCTION_LIKE_LOGS` env var to all apps and so it's safe to merge the latest version of this gem into an app like `signon`?

@floehopper Just saw this now, my apologies. Thank you for confirming this with someone else and bumping the version accordingly. 🙏🏻

Also, we identified a small issue that affects development, you might want to bump the version again. See [PR here](https://github.com/alphagov/govuk_app_config/pull/309).

PR #303, opened 2023-07-03
Add support Open Telemetry instrumentation in Rails

theseanything:
This adds the ability to enable Open Telemetry instrumentation in Rails by setting the "ENABLE_OPEN_TELEMETRY" environment variable to "true".

The service name is automatic set to the Rails application name.

---
This is work done part as GIFT week and should initially only be enabled in integration. Should be removed if ultimately not implemented/used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 8.1.0
+
+* Add ability to enable Open Telemetry instrumentation for Rails applications.
+
 # 8.0.2
 
 * Fix the ability to collect Sidekiq metrics in GovukPrometheusExporter. ([#299](https://github.com/alphagov/govuk_app_config/pull/299))
--- a/README.md
+++ b/README.md
@@ -104,6 +104,9 @@ end
 
 `GovukError.configure` has the same options as the Sentry client, Raven. See [the Raven docs for all configuration options](https://docs.sentry.io/clients/ruby/config).
 
+## Open Telemetry
+
+To enable Open Telemetry instrumentation for Rails set the ENABLE_OPEN_TELEMETRY="true" environment variable.
 
 ## Prometheus monitoring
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 8.1.0
+
+* Add ability to enable Open Telemetry intrumentation for Rails applications.
```
sengi, line 3: tiniest of nits:
```suggestion
* Add ability to enable Open Telemetry instrumentation for Rails applications.
```

README.md
```diff
@@ -104,6 +104,9 @@ end
 
 `GovukError.configure` has the same options as the Sentry client, Raven. See [the Raven docs for all configuration options](https://docs.sentry.io/clients/ruby/config).
 
+## Open Telemetry
+
+To enable Open Telemetry intrumentation for Rails set the ENABLE_OPEN_TELEMETRY="true" environment variable.
```
sengi, line 109:
```suggestion
To enable Open Telemetry instrumentation for Rails set the ENABLE_OPEN_TELEMETRY="true" environment variable.
```

lib/govuk_app_config/railtie.rb
```diff
@@ -8,6 +8,10 @@ class Railtie < Rails::Railtie
       end
     end
 
+    initializer "govuk_app_config.configure_open_telemetry" do |app|
+      GovukOpenTelemetry.configure(app.class.module_parent_name.underscore)
```
sengi, line 12: You've probably already thought of this, but you'll need to make the service name configurable at some point if this gets used for real, otherwise things like the draft- deployments will be hard to distinguish from the regular ones etc. No problem for now though.

sengi (approved):

lib/govuk_app_config/railtie.rb
theseanything: Yup good point! Also this won't work for our non-rails apps e.g. search api

PR #306, opened 2023-07-05
Fix dropped logs due to incorrect error field type

yuetylauiris:
Kibana validates log lines by index patterns and "error" is supposed to be an object. Since error is a string in logstaher's world we are renaming it to "message" which will be accepted as string.

### DIFF
--- a/lib/govuk_app_config/govuk_json_logging.rb
+++ b/lib/govuk_app_config/govuk_json_logging.rb
@@ -50,6 +50,11 @@ module GovukJsonLogging
     # Elasticsearch index expect source to be an object and logstash defaults
     # source to be the host IP address causing logs to be dropped.
     Rails.application.config.logstasher.source = {}
+    # Elasticsearch index expect error to be an object and logstash defaults
+    # error to be a string causing logs to be dropped.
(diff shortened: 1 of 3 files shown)


tunylund (approved):

lib/govuk_app_config/govuk_json_logging.rb
```diff
@@ -50,6 +50,11 @@ def self.configure
     # Elasticsearch index expect source to be an object and logstash defaults
     # source to be the host IP address causing logs to be dropped.
     Rails.application.config.logstasher.source = {}
+    # Elasticsearch index expect error to be an object and logstash defaults
+    # error to be a string causing logs to be dropped.
+    Rails.application.config.logstasher.field_renaming = {
+      error: :message,
```
theseanything, line 56: Did you mean this to be "error" to "error.message"? This appears to renaming the field "error" to "message".

kevindew:
We've not changed anything logstasher though have we? So were these items being dropped before this change?

It does seem to lose meaning if we rename error to message, do we instead need to fix our logit configuration to not expect a json object for a message?

PR #311, opened 2023-07-31
GovukAppConfig silences OpenTelemetry log output when running a rake task

KludgeKML:
https://github.com/alphagov/govuk_app_config/pull/307 removes OpenTelemetry initialisation in the console to avoid flooding the console with OpenTelemetry startup messages.

This PR adds the additional behaviour that when OpenTelemetry _is_ configured (ie we are not on a console), we do a second check to see whether we're in a rake task and silence OpenTelemetry log messages so that Rake tasks with simple text output are not swamped with OpenTelemetry startup messages.

Since we're already assuming Rails.const_defined? is available, we use it to check for Rake, then call Rake.application.top_level tasks, which returns an empty array if running in rails, but a list of tasks if running through rake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* GovukAppConfig silences OpenTelemetry log output when running a rake task ([#311](https://github.com/alphagov/govuk_app_config/pull/311))
+
 # 9.0.4
 
 * Fix an issue with Rails.logger being not an instance of ActiveSupport::Logger. Rails expects Rails.logger to have methods that Ruby STD Logger does not provide. e.g. `silence()` ([#309](https://github.com/alphagov/govuk_app_config/pull/309))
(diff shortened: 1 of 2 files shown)


robinjam:
It looks like we could configure it not to log anything (but I'm not sure what else it logs - could be important stuff)

```ruby
OpenTelemetry::SDK.configure do |c|
  c.logger = Logger.new(File::NULL)
end
```

https://www.rubydoc.info/gems/opentelemetry-sdk/OpenTelemetry/SDK

KludgeKML:
@robinjam / @theseanything I've updated this PR to try James's suggestion, will try it on a branch of email-alert-api to confirm it works as expected, then if that goes okay I'll tag you both in as reviewers.

theseanything (approved):

KludgeKML:
@theseanything Since you approved I've moved the changed code into lib/govuk_app_config/govuk_open_telemetry.rb instead of it being in the railtie (so that the OpenTelemetry::SDK config code is all in one block). That still okay?

PR #314, opened 2023-09-05
Update warning about address in use for Prometheus metrics server

theseanything:
This makes the warning message less confusing, as error is about the Prometheus metrics server failing to start rather than being unable to connect to Prometheus itself.


KludgeKML (approved):
LGTM

PR #316, opened 2023-09-08
Add ability to specify custom collectors

theseanything:
This extends application Prometheus exporter configuration to load custom collectors. This is required by application that would like to support global metrics (e.g. user/article counts). Specifically this is need to support Signon provide metrics for token expiry timestamps.

Following the global metrics pattern outlined in the [prometheus_exporter docs](https://github.com/discourse/prometheus_exporter#global-metrics-in-a-custom-type-collector).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * GovukAppConfig silences OpenTelemetry log output when running a rake task ([#311](https://github.com/alphagov/govuk_app_config/pull/311))
 * Update warning message for Prometheus metric server address already in use.
+* Add ability to support custom collectors in the Prometheus exporter.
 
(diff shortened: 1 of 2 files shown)


samsimpson1 (approved):

sengi (commented):
LGTM. I'm guessing the need for a custom collector comes about because it's the least-hassle way (given discourse/prometheus-exporter) to have one process responsible for updating global metrics in a multi-process app? Or am I way off? 😅

sengi:
In any case, probably worth having a skim through https://prometheus.io/docs/instrumenting/writing_exporters/#collectors if you haven't already (dunno how much of it will apply here but might be useful background since writing custom collectors is a little bit off the beaten track)

PR #320, opened 2023-10-03
Get controller labels from controller, not params

richardTowers:
I've got an open PR to do this for the default default in prometheus_exporter itself:

`discourse/prometheus_exporter/pull/293`

But it's unclear whether the maintainers will accept that, and if so how long that will take.

In the meantime, we've got quite a bit of noise in our metrics because it's currently possible for an HTTP request to overwrite the action / controller labels by passing them in as request parameters like this:

    curl -v http://127.0.0.1:3000/ --data 'controller=test'

This commit extends the default middleware so we have more sensible defaults for both Rails and Sinatra apps. I couldn't work out any good labels for sinatra apps that wouldn't be potentially high cardinality, so those are just blank.

### DIFF
--- a/lib/govuk_app_config/govuk_prometheus_exporter.rb
+++ b/lib/govuk_app_config/govuk_prometheus_exporter.rb
@@ -4,6 +4,44 @@ require "prometheus_exporter/server"
 require "prometheus_exporter/middleware"
 
 module GovukPrometheusExporter
+
+  #
+  # See https://github.com/discourse/prometheus_exporter/pull/293
+  #
+  # RailsMiddleware can be removed and replaced with the default middleware if
+  # that PR is merged / released
+  #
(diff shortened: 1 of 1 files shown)


sihugh (approved):
Feels a bit surprising to get something from `env` and then access object properties directly rather than accessing a hash/array/string. I guess it works though?

richardTowers:
> Feels a bit surprising to get something from env and then access object properties directly rather than accessing a hash/array/string. I guess it works though?

Yeah, it's a bit weird, but some of the values on the `env` hashes are full on object instances. [`action_controller.instance` is the current controller's instance](https://apidock.com/rails/v6.1.3.1/ActionDispatch/Request/controller_instance).

I've tested it locally (as this isn't covered by any tests in the gem 😬), and it does work in frontend.

PR #325, opened 2023-10-09
Allow gov.uk domains to embed pages

AgaDufrat:
We initially added the strict and OWASP recommended 'none' directive based on the assumption that only side-by-side-browser tool (retired in November 2022) was preventing us from implementing it.

However some other internal GOV.UK apps use iframes:
- Search Admin 
  - Best bets (queries) 
  - External links (recommended-links)
- Content Publisher (Preview feature)

This policy will still ensure sufficient security yet will allow internal GOV.UK domains to embed pages.  It's added to the global base policy because given the number of frontend application it may be difficult to predict which frontend app renders the page that we want to iframe. It will reduce the need to apply a CSP modification in individual apps.

[Trello](https://trello.com/c/QC9k1z4a/3303-prevent-iframing-of-govuk-3)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.5.0
+
+* Allow gov.uk domains to embed pages in the global Content Security Policy ([#325](https://github.com/alphagov/govuk_app_config/pull/325))
+
 # 9.4.0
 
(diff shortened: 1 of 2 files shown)


robinjam (approved):

ChrisBAshton (commented):
Nice job 🚀

PR #327, opened 2023-11-15
Enable adding custom LogStasher fields from apps

mike29736:
I.e. allow apps to add their own custom fields in addition to those that
`GovukJsonLogging` already adds.

It looks like `LogStasher.add_custom_fields` can only be called one
time, otherwise subsequent calls overwrite previous ones.

We found this [happening in the wild in the Content Store app](https://github.com/alphagov/content-store/blob/v33/config/initializers/logstasher.rb#L2), where
`govuk_request_id` et al were missing from the Rails framework logs but
present whenever the logger had been invoked directly from our own code.

This approach does solve the problem for Content Store (along with a
change to Content Store to actually use this option), but I'm keen to hear
suggestions of alternatives.

(https://trello.com/c/jDSDGNyn/814-measure-and-record-our-publishing-latency-sli)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Enable adding custom LogStasher fields from apps ([#327](https://github.com/alphagov/govuk_app_config/pull/327))
+
 # 9.6.0
 
 * Allow YouTube thumbnails from https://i.ytimg.com in the global Content Security Policy ([#328](https://github.com/alphagov/govuk_app_config/pull/328))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 9.7.0
 
 * Enable adding custom LogStasher fields from apps ([#327](https://github.com/alphagov/govuk_app_config/pull/327))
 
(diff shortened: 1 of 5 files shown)


aldavidson:
I believe this works, but it would be nice if there was a way to test that it does in the specs. Is that possible?

aldavidson:
> I believe this works, but it would be nice if there was a way to test that it does in the specs. Is that possible?

After actually trying & failling to do this, I don't think it's realistically possible to replicate the scenario this is configured for in a test case. The custom fields are added via LogStasher, which uses log subscription callbacks to attach to Rails apps, and Thread.current values to store the custom fields - which means that to test this, we'd need to couple very tightly to LogStasher implementation.  While LogStasher itself [does something to address this in its own specs](https://github.com/shadabahmed/logstasher/blob/0d7e3d1e79620d837accb6f111758d24c61b6b49/spec/integration_spec.rb), I don't think it's right for us to do it here.

aldavidson (approved):

robinjam (approved):
LGTM 👍

mike29736:
Thanks for the reviews :bow:

I'll resolve the merge conflict, bump the gem a minor version number (in an additional commit) and merge this when I'm back at work on Monday

aldavidson:
I eventually found a way of testing this that's only medium-awkward. I've submitted it as a separate PR (#330) based off this branch, so you can accept or not as you see fit.

PR #330, opened 2023-11-24
Add RSpec examples to test the adding of custom fields via LogStasher

aldavidson:
As [previously discussed](https://github.com/alphagov/govuk_app_config/pull/327#issuecomment-1824376665), the way LogStasher hooks into Rails makes it difficult to test that configured custom fields actually get added. ([Trello card](https://trello.com/c/jDSDGNyn/814-measure-and-record-our-publishing-latency-sli))

After some trial and error, I found a way of testing it with only a bit of awkward coupling.

### DIFF
--- a/spec/lib/govuk_json_logging_spec.rb
+++ b/spec/lib/govuk_json_logging_spec.rb
@@ -3,6 +3,11 @@ require "rails"
 require "govuk_app_config/govuk_json_logging"
 require "rack/test"
 
+class TestController < ActionController::Base
+  include LogStasher::ActionController::Instrumentation
+  def index; end
+end
+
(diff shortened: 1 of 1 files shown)


deborahchua (approved):

mike29736:
Thanks for this @aldavidson, and thanks for reviewing it @deborahchua :bow: 

I've cherry-picked this commit on my local machine and I'm going to flesh out the commit message a bit to explain what `append_info_to_payload` and `logstasher_add_custom_fields_to_payload` are before pushing it up to [my open pull request](https://github.com/alphagov/govuk_app_config/pull/327). So, I'll close this now

PR #334, opened 2023-12-20
Add GOVUK domains to script src CSP

AshGDS:
The `assets.publishing.service.gov.uk` domain can't POST to the feedback form without this, so users can't report problems with our CSV preview pages.


AgaDufrat (approved):
Thanks!
Feel free to release the gem in the same PR.

PR #336, opened 2023-12-21
Revert "Add GOVUK domains to script src CSP"

AshGDS:
This reverts commit b30410cbfd7c72cec77e6580e13147ed55d535be.

We thought it would solve a CORS issue, but it did not so we don't need this change.

### DIFF
--- a/lib/govuk_app_config/version.rb
+++ b/lib/govuk_app_config/version.rb
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "9.8.0".freeze
+  VERSION = "9.8.1".freeze
 end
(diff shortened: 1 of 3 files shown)


lib/govuk_app_config/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukAppConfig
-  VERSION = "9.8.0".freeze
+  VERSION = "9.9.0".freeze
```
ChrisBAshton, line 2: Nitpick: we could make this `9.8.1`, because it's not a breaking change (insofar as no other apps will have made any changes in light of the change introduced in 9.8.0) and it's more of a bugfix than a new feature.

ChrisBAshton (approved):
Approved, but suggest we downgrade to 9.8.1 (in version.rb and CHANGELOG.md)

lib/govuk_app_config/version.rb
AshGDS: Thanks @ChrisBAshton, will update :+1:

PR #343, opened 2024-01-25
Use shorthand lambda syntax

### LINKED ISSUE #341: Fails on Ruby 3.3 with `lambda method requires a literal block`.
This prevents consumers of govuk_app_config from building with Ruby 3.3 or above.

See [Kernel#lambda raises when passed Proc instance](https://rubyreferences.github.io/rubychanges/3.3.html#kernellambda-raises-when-passed-proc-instance) in the 3.3 release notes.

```
ArgumentError: the lambda method requires a literal block (ArgumentError)

    ::Kernel.instance_method(m).bind_call(self, *args, &block)
                                            ^^^^^^^^^^^^^^^^^^^
vendor/bundle/ruby/3.3.0/gems/govuk_app_config-9.8.1/lib/govuk_app_config/govuk_error/configuration.rb:88:in `ignore_excluded_exceptions_in_data_sync'
vendor/bundle/ruby/3.3.0/gems/govuk_app_config-9.8.1/lib/govuk_app_config/govuk_error/configuration.rb:70:in `set_up_defaults'
vendor/bundle/ruby/3.3.0/gems/govuk_app_config-9.8.1/lib/govuk_app_config/govuk_error/configuration.rb:12:in `initialize'
```

Repro: take pretty much any GOV.UK app (e.g. alphagov/release), update its .ruby-version to 3.3.0 and run the project's tests.

robinjam:
Fixes #341

`SimpleDelegator` was intercepting the `lambda` call via `method_missing`, and delegating to `Kernel#lambda`, passing the given block. This explodes on Ruby 3.3 because `lambda` only accepts block literals now.

Replacing the `lambda` call with this shorthand syntax sidesteps the issue. RuboCop is unhappy because it prefers the old syntax for multiline lambdas, so I've had to disable that cop.

Tested locally against Content Data Admin.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,5 @@
 inherit_gem:
   rubocop-govuk:
     - config/default.yml
+Style/Lambda:
+  Enabled: false
--- a/README.md
+++ b/README.md
@@ -96,9 +96,9 @@ and occurs at the time of a data sync, then it will be excluded even if the cust
 
 ```ruby
 GovukError.configure do |config|
-  config.before_send = lambda do |event, hint|
+  config.before_send = ->(event, hint) {
     hint[:exception].is_a?(ErrorWeWantToIgnore) ? nil : event
-  end
+  }
 end
 ```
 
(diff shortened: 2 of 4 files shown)


sengi:
I'm pretty suspicious of the use of SimpleDelegator here but I suspect pulling on that string might lead to a much bigger refactor — so I think your fix is the right one at the right time.

robinjam:
> I'm pretty suspicious of the use of SimpleDelegator here but I suspect pulling on that string might lead to a much bigger refactor — so I think your fix is the right one at the right time.

Agreed, I'm always a little wary of `SimpleDelegator` and it certainly feels like a code smell here. Something to look into in the future

PR #355, opened 2024-02-27
Add single cookie consent api URLs

andysellick:
## What / why
Adds the URLs for the single cookie consent API into the CSP. Once the single consent API is enabled on GOV.UK the JS will be making XMLHttprequests to the staging and production environments for the single consent api, so these URLs need to be added to the CSP.

Trello card: https://trello.com/c/dmlTAzKB/140-implement-single-consent-api

### DIFF
--- a/lib/govuk_app_config/govuk_content_security_policy.rb
+++ b/lib/govuk_app_config/govuk_content_security_policy.rb
@@ -71,7 +71,9 @@ module GovukContentSecurityPolicy
                        *GOVUK_DOMAINS,
                        *GOOGLE_ANALYTICS_DOMAINS,
                        # Speedcurve real user monitoring (RUM) - as per: https://support.speedcurve.com/docs/add-rum-to-your-csp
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Looks good to me. Only thing I wasn't sure about was whether to include the `https://` protocol, as the speedcurve link doesn't have that. The documentation suggests it's okay to include that though - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources

andysellick:
@richardTowers that's a good point, have updated, thanks!

kevindew:
For future reference, it's slightly better to include https as it means it will only connect on that protocol. Without it will allow non-TLS http.

andysellick:
Thanks @kevindew I hadn't realised that. Should I go back and change this now?

kevindew:
Nah don't worry. I expect at some point in the future, we'll (or I'll) do a blanket update of them.

PR #361, opened 2024-03-26
9.10.0: Always initialise GovukPrometheusExporter.

sengi:
Some client apps have started to assume that the Prometheus "exporter" process (i.e. the process that aggregates counters in multi-process apps) is always available, including in tests.

Now that we handle the case where the `/metrics` port is already taken, there's really no reason not to initialise GovukPrometheusExporter.

Remove the confusing heuristics in `GovukPrometheusExporter.should_configure` and just always attempt to initialise unless we're running under `rails console`. When the init fails, we just print a warning and continue anyway, so this should be a net win for robustness.

If we find edge cases where this still needs to be configurable then I propose we address those if/when we find and understand them. I expect we won't need to, though.

Kudos to @leenagupte for reporting the issue.


KludgeKML (approved):
LGTM, nice simplification.

sengi:
This should eliminate the need for the workaround in https://github.com/alphagov/finder-frontend/pull/3318 and https://github.com/alphagov/search-api-v2/pull/246 (hopefully without introducing any other significant issues 😅)

sengi:
Realised I'd missed out the words "environment variable" in the changelog entry 🙃 (force-pushed to update)

PR #366, opened 2024-04-25
Add GDS::SSO::PermissionDeniedError to excluded exceptions list

AlexAvlonitis:
Related to: https://github.com/alphagov/gds-sso/pull/300

-  We do not need to capture these errors generated from gds-sso. They do not indicate that something went wrong with the application; rather, they indicate that a user does not have permissions to access a resource.

trello card: https://trello.com/c/qb1qpwnr/1369-additional-functionality-to-gds-sso-gem-for-intercepting-401s-and-providing-a-route-constraint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.11.0
+
+* Add GDS::SSO::PermissionDeniedError to excluded exceptions list ([#366](https://github.com/alphagov/govuk_app_config/pull/366))
+
 # 9.10.0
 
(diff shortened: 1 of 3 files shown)


kevindew (commented):
Change looks good. It would be good if the commit explained why we don't want this send to error to Sentry.

Would you be able to add an entry to the changelog too?

kevindew:
> Thanks for the review @kevindew , I updated the things you suggested, I followed the pattern for the changelog message and version (I noticed it does not have unreleased section), should I update the gem's version as part of this PR?

We normally just add an unreleased section each time a new one is added.

It's no biggie though to update the gem version as part of this PR since this is a relatively slow moving gem.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 9.10.1
```
kevindew, line 1:
```suggestion
# 9.11.0
```

This isn't a bug fix, so should be a minor version update. As per: https://semver.org/

kevindew (approved):
Looks good to me. Lets just check with the owning team: Platform Security Reliability team before merging. Would you mind dropping them a message on Slack (#govuk-platform-security-reliability-team)

MuriloDalRi (approved):

PR #372, opened 2024-05-21
Fix OpenTelemetry span processing when using logstasher

theseanything:
The logstasher gem monkey patches the `process_action.action_controller` instrumentation. However the gem hasn't been updated for Rails 7, which assumes that the payload has the 'request' object available. This request object is used by the OpenTelemetry SDK to process spans. This generates lots of error logs from the SDK, and prevents us from recording some spans.

This re-monkey patches the instrumentation to use a raw_payload that is expected in Rails 7.  Ideally, this would be fixed in the logstasher gem however there has been little active development over the last year. Longer term fix might be to move to a maintained logging library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Fix OpenTelemetry errors when using with Logstasher gem ([#372](https://github.com/alphagov/govuk_app_config/pull/372))
+
 # 9.11.0
 
 * Add GDS::SSO::PermissionDeniedError to excluded exceptions list ([#366](https://github.com/alphagov/govuk_app_config/pull/366))
--- a/lib/govuk_app_config/govuk_json_logging.rb
+++ b/lib/govuk_app_config/govuk_json_logging.rb
@@ -16,6 +16,12 @@ module GovukJsonLogging
   end
 
   def self.configure(&block)
+    # Fixes the monkey patch from the logstasher gem to support Rails 7
+    config = Rails.application.config.logstasher
+    if (!config.controller_monkey_patch && config.controller_monkey_patch != false) || config.controller_monkey_patch == true
+      require_relative "./govuk_json_logging/rails_ext/action_controller/metal/instrumentation"
+    end
+
     configuration = Configuration.new
 
     configuration.instance_eval(&block) if block_given?
--- a/lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
+++ b/lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
@@ -1,0 +1,65 @@
+# This is copied from
+# https://github.com/shadabahmed/logstasher/blob/main/lib/logstasher/rails_ext/action_controller/metal/instrumentation.rb
+# Changes have been highlight in comments, otherwise the code is the same.
+
+module ActionController
+  module Instrumentation
+    alias_method "orig_process_action", "process_action"
+
+    def process_action(*args)
+      # The raw payload has been updated to reflect the payload structure used
+      # in Rails 7.1, primarily the addition of the `headers`, `request` keys
+      # and using `request.filtered_path` instead of `request.fullpath`.
+      # https://github.com/rails/rails/blame/d39db5d1891f7509cde2efc425c9d69bbb77e670/actionpack/lib/action_controller/metal/instrumentation.rb#L60
+      raw_payload = {
+        controller: self.class.name,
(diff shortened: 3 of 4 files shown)


lib/govuk_app_config/govuk_json_logging.rb
```diff
@@ -16,6 +16,12 @@ def add_custom_fields(&block)
   end
 
   def self.configure(&block)
+    # Fixes the monkey patch from the logstasher gem to support Rails 7
+    config = Rails.application.config.logstasher
+    if (!config.controller_monkey_patch && config.controller_monkey_patch != false) || config.controller_monkey_patch == true
+      require_relative "./govuk_json_logging/rails_ext/action_controller/metal/instrumentation"
+    end
+
```
theseanything, line 24: This is a copy pasta from the logstasher gem to keep the conditional require:
https://github.com/shadabahmed/logstasher/blob/0d7e3d1e79620d837accb6f111758d24c61b6b49/lib/logstasher.rb#L110

richardTowers: 😅  what an if condition...

is there no way to refer to this module within the logstasher gem, without vendoring it?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Fix OpenTelemetry errors when using with Logstasher gem ([#372](https://github.com/alphagov/govuk_app_config/pull/372))
```
ChrisBAshton, line 3: Ahhh thank you - I was just thinking about this error today and how it would be good to figure out how to get rid of it!

lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
...
+          payload[key] = value
+        end
+
+        LogStasher.request_context.each do |key, value|
+          payload[key] = value
+        end
+        result
+      end
+    end
+    alias_method "logstasher_process_action", "process_action"
```
ChrisBAshton, line 54: Any reason why this is declared down here and not just below the equivalent call on line 3?

lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
+module ActionController
+  module Instrumentation
+    alias_method "orig_process_action", "process_action"
+    def process_action(*args)
```
ChrisBAshton, line 4: Is this copy-pasted from anywhere? If so, please add the link - and if not, this might be one of those rare occasions that actually benefits from interspersing more inline comments explaining what you're doing in each sequence of steps 🙏

lib/govuk_app_config/govuk_json_logging.rb
theseanything: ah sorry, not sure what you mean?

lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
theseanything: Added comments to highlight whats changed. The only change is updating the raw_payload to be similar to whats used in Rails 7.1.

lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
theseanything: This was copied over from the original patch.

lib/govuk_app_config/govuk_json_logging/rails_ext/action_controller/metal/instrumentation.rb
ChrisBAshton: Ah ok - thanks 👍

ChrisBAshton (approved):

lib/govuk_app_config/govuk_json_logging.rb
richardTowers: I meant if the code in `instrumentation` was the same as https://github.com/shadabahmed/logstasher/blob/0d7e3d1e79620d837accb6f111758d24c61b6b49/lib/logstasher/rails_ext/action_controller/metal/instrumentation.rb, was there some way to refer to it that doesn't involve copying it into this gem. But I can see from the comments you've added that it is actually different, so no worries!

PR #381, opened 2024-06-20
Set time_zone to London in all GOV.UK apps

richardTowers:
Our conventions for rails apps say that we should set time_zone to London:

https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#specify-london-as-the-timezone

In practice however, only about half of the apps do so. Based on the apps running in integration:

Apps which set the time_zone to London:

    AccountApi
    Collections
    CollectionsPublisher
    ContentPublisher
    Collections
    Frontend
    GovernmentFrontend
    Static
    EmailAlertApi
    Frontend
    GovernmentFrontend
    GovukChat
    LocationsApi
    Maslow
    PlacesManager
    Publisher
    ReleaseApp
    SearchApiV2
    Signon
    SmartAnswers
    SpecialistPublisher
    Static
    Whitehall

Apps which leave the time_zone as it's default UTC:

    AuthenticatingProxy
    Contacts
    ContentDataAdmin
    ContentPerformanceManager
    ContentStore
    ContentTagger
    ContentStore
    RouterApi
    EmailAlertFrontend
    Feedback
    FinderFrontend
    HmrcManualsApi
    LinkCheckerApi
    LocalLinksManager
    ManualsPublisher
    RouterApi
    SearchAdmin
    SearchV2Evaluator
    ServiceManualPublisher
    ShortUrlManager
    Support
    SupportApi
    Transition
    TravelAdvicePublisher

Having the apps use inconsistent time_zones feels like a recipe for trouble.

`config.time_zone` has to be set before `active_support.initialize_time_zone` runs, so we specify that our initializer must run before it. Unfortunately this means that there's no way for applications to set their own value of `time_zone`. I think this is probably desirable for consumers of `govuk_app_config` though - I can't think of a legitimate reason why we'd want some apps using different time zones.

Technically this is a breaking change, as it could change behaviour in apps which are using timezone-specific methods and expecting to get UTC times out. Hopefully we're already explicit about when we want UTC and when we want local time though.

Once this is merged, I'll remove the bit in the docs about setting time_zone, as it will be automatic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Adds the basics of a GOV.UK application:
 - Statsd client for reporting stats (deprecated; use Prometheus instead)
 - Rails logging
 - Content Security Policy generation for frontend apps
+- Sets the time zone to London
 
 ## Installation
 
@@ -181,6 +182,9 @@ GovukContentSecurityPolicy.configure
 
 Some frontend apps support languages that are not defined in the i18n gem. This provides them with our own custom rules for these languages.
 
+## Time zone
+
+This gem sets `config.time_zone` to `"London"` - this cannot currently be overridden in application config.
 
 ## License
 
--- a/lib/govuk_app_config.rb
+++ b/lib/govuk_app_config.rb
@@ -9,5 +9,6 @@ require "govuk_app_config/govuk_prometheus_exporter"
 if defined?(Rails)
   require "govuk_app_config/govuk_json_logging"
   require "govuk_app_config/govuk_content_security_policy"
+  require "govuk_app_config/govuk_timezone"
   require "govuk_app_config/railtie"
 end
--- a/lib/govuk_app_config/govuk_timezone.rb
+++ b/lib/govuk_app_config/govuk_timezone.rb
@@ -1,0 +1,5 @@
+module GovukTimezone
+  def self.configure(config)
+    config.time_zone = "London"
+  end
+end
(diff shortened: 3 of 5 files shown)


MuriloDalRi (dismissed):

AgaDufrat (approved):
Seems sensible to standardise this in all the apps.

A scenario where it could be an issues is if we depend on some external process that runs in UTC but I also can’t think of where it would be an issue within application.

Irrelevant (because we are using cron schedule for it) example: 
In Content Data API an import has to run (currently 40 minutes) after a GoogleAnalytics query has run and GA schedule is in UTC. 

The [Rails/TimeZone](https://www.rubydoc.info/gems/rubocop/0.57.2/RuboCop/Cop/Rails/TimeZone) cop should also keep things in check.

richardTowers:
👍🏻  Thanks Aga and Murilo.

> A scenario where it could be an issues is if we depend on some external process that runs in UTC but I also can’t think of where it would be an issue within application.

Yeah - it's also still possible to refer to UTC times explicitly, if that's what the code needs. Which we should be doing anyway hopefully. If we wanted to do like a sidekiq schedule thing (which I appreciate isn't our preferred approach, instead favouring k8s cronjobs) that had to run daily at a particular UTC time, that would still be fine.

Having thought about it a little more, I was worried about the possibility of this causing confusion if someone _is_ trying to set time_zone to something else explicitly, so I've made it log what its doing and error in the situation where someone's trying to do something the gem is going to undo. This meant I had to change the tests quite a bit, so I think it'll need a re-review I'm afraid.

MuriloDalRi (approved):

MuriloDalRi:
@richardTowers LGTM!

PR #383, opened 2024-06-24
Exclude Slimmer::SourceWrapperNotFoundError

unoduetre:

## What

We keep getting the following exception in production: https://govuk.sentry.io/issues/4958108827/?environment=production&environment=production-eks&environment=producton&project=-1&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=9

After [the discussion](https://gds.slack.com/archives/C052X8S2P8A/p1718796354165479) there was a suggestion to keep the SourceWrapperNotFoundError instead of reverting the PRs introducing it. The error should not be sent to Sentry, but developers should still receive it in their local environments and in the tests.

There is a separate, related [PR](https://github.com/alphagov/slimmer/pull/334) applying changes to slimmer.

## Why

[Trello ticket](https://trello.com/c/iyPxdU09/2495-sentry-errors-investigate-fix-frontend-nomethoderror)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 9.13.0
 
 * Add Slimmer::SourceWrapperNotFoundError to excluded exceptions.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Add Slimmer::SourceWrapperNotFoundError to excluded exceptions.
+
 # 9.12.0
 
 * Set time_zone to London in all GOV.UK apps ([#381](https://github.com/alphagov/govuk_app_config/pull/381))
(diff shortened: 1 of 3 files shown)


hannako:
> The error should not be sent to Sentry, but developers should still receive it in their local environments and in the tests.

Would it also still be visible in Kibana?

unoduetre:
> > The error should not be sent to Sentry, but developers should still receive it in their local environments and in the tests.
> 
> Would it also still be visible in Kibana?

Yes, I've checked on integration and the error is sent to Kibana, but not to Sentry.

hannako (approved):

MuriloDalRi:
Could you please add a commit to release a new version of the gem to this PR to avoid us having to review it twice?

MuriloDalRi (approved):

PR #392, opened 2024-07-22
Configure time_zone with govuk_time_zone

richardTowers:
In the original implementation of the GovukTimezone module we assumed that literally all GOV.UK apps would use the London timezone. We didn't really consider how to make it easy for apps to override this if needed.

Since then, we've found a couple of examples where apps genuinely do want to override the timezone:

- publishing-api, which always wants to present dates as UTC because its an API
- travel-advice-publisher, which always wants to present dates as UTC because a majority of its users are in non-London timezones (so BST is confusing)

Currently, these apps override GovukTimezone by inserting an initializer between it and active_support.initialize_time_zone like this:

    initializer "publishing_api.configure_timezone",
                after: "govuk_app_config.configure_timezone",
                before: "active_support.initialize_time_zone" do
      config.time_zone = "UTC"
    end

    # https://github.com/alphagov/publishing-api/blob/252f47a1/config/application.rb#L123-L129

This is a bit of a faff, and it also results in some very confusing logs.

Because the GovukTimezone still runs in this setup, it would log "govuk_app_config changing time_zone from UTC (the default) to London" ... but then we'd set it back to UTC without logging anything, making it seem from the logs that the time_zone was London when it was actually UTC. Bleurgh.

Unfortunately, it's not possible to work out just from config.time_zone whether the time_zone is set to UTC because its the default (in which case we should override it to London), or because the app has explicitly set it (in which case we should leave it alone).

I think the best approach is to introduce a new method govuk_time_zone, which won't be set by default to allow us to distinguish these scenarios.

This has the nice benefit that we can be a bit more strict about what's an error case and what just requires a log message.

If config.time_zone is set to anything other than UTC, that's an error. If config.govuk_time_zone is set, we use that, otherwise we override UTC to London and log a message.

This will be slightly annoying to release, because we'll have to remove all the instances of config.time_zone = "London" across our apps. If we want to be explicit, we can set config.govuk_time_zone. If we're happy with the default, then we can just remove the line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.14.0
+
+* Configure `time_zone` with `govuk_time_zone` [#392](https://github.com/alphagov/govuk_app_config/pull/392)
+
 # 9.13.1
 
 * Update dependencies
--- a/README.md
+++ b/README.md
@@ -184,7 +184,17 @@ Some frontend apps support languages that are not defined in the i18n gem. This 
 
 ## Time zone
 
-This gem sets `config.time_zone` to `"London"` - this cannot currently be overridden in application config.
+This gem sets `config.time_zone` to `"London"` by default.
+
+If you require a different time zone, you can set it with `config.govuk_time_zone`:
+
+```ruby
+config.govuk_time_zone = "UTC"
+```
+
+Note that we've introduced a new config field (`govuk_time_zone`) here, because
+it's otherwise not possible to distinguish between an app using UTC as the
+default and an app explicitly asking for UTC.
 
 ## License
 
--- a/lib/govuk_app_config/govuk_timezone.rb
+++ b/lib/govuk_app_config/govuk_timezone.rb
@@ -1,14 +1,12 @@
 module GovukTimezone
   def self.configure(config)
-    case config.time_zone
-    when "UTC"
-      Rails.logger.info "govuk_app_config changing time_zone from UTC (the default) to London"
-    when "London"
-      Rails.logger.info "govuk_app_config always sets time_zone to London - there is no need to set config.time_zone in your app"
-    else
-      raise "govuk_app_config prevents configuring time_zones other than London - config.time_zone was set to #{config.time_zone}"
-    end
+    raise "govuk_app_config prevents configuring time_zone with config.time_zone - use config.govuk_time_zone instead" unless config.time_zone == "UTC"
 
-    config.time_zone = "London"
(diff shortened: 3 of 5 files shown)


kevindew (approved):
Looks good to me 👍 Thanks

Have you given it a try in a couple of apps just to rule out any surprises?

richardTowers:
Works in publishing-api:

```
 ~ ▶ govuk ❯ publishing-api ▶ main ▶ 3✎ ▶ 2⚑ ▶ % ▶ git diff config
diff --git a/config/application.rb b/config/application.rb
index 132c34ae..6bcaebb8 100644
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,8 +33,8 @@ module PublishingAPI
     # These settings can be overridden in specific environments using the files
     # in config/environments, which are processed later.
     #
-    # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+    config.govuk_time_zone = "UTC"

     # Only loads a smaller set of middleware suitable for API only apps.
     # Middleware like session, flash, cookies can be added back manually.
@@ -119,13 +119,5 @@ module PublishingAPI
     )

     config.debug_exception_response_format = :api
-
-    # Even though most GOV.UK apps use the London time_zone, publishing-api almost always wants to
-    # present times in UTC, so we set that as the default time_zone
-    initializer "publishing_api.configure_timezone",
-                after: "govuk_app_config.configure_timezone",
-                before: "active_support.initialize_time_zone" do
-      config.time_zone = "UTC"
-    end
   end
 end
 ~ ▶ govuk ❯ publishing-api ▶ main ▶ 3✎ ▶ 2⚑ ▶ % ▶ rails runner 'puts Time.zone'
(GMT+00:00) UTC
```

In government-frontend (which now incorrectly sets config.time_zone):

```
 ~ ▶ govuk ❯ government-frontend ▶ main ▶ 2✎ ▶ ERROR ▶ % ▶ rails runner 'puts Time.zone'
/Users/richard.towers/govuk/govuk_app_config/lib/govuk_app_config/govuk_timezone.rb:3:in `configure': govuk_app_config prevents configuring time_zone with config.time_zone - use config.govuk_time_zone instead (RuntimeError)
```

Fixing it to set config.govuk_time_zone instead:

```
~ ▶ govuk ❯ government-frontend ▶ main ▶ 3✎ ▶ % ▶ rails runner 'puts Time.zone'
(GMT+00:00) London
```

PR #399, opened 2024-10-08
Update SidekiqRedis healthcheck to work with Sidekiq 7

davidgisbey:
This healthcheck has been failing for applications that have upgraded to Sidekiq 7. Sidekiq 7 uses `redis-client`' which has a different API to `redis` which was used in Sidekiq 6.

To access `redis_info` you now have to call

```
Sidekiq.default_configration.redis_info
```

while in Sidekiq 6 you call

```
Sidekiq.redis_info
```

This updates the healthcheck to work with both versions of Sidekiq checking if`Sidekiq responds to `.redis_info`.

If it does it calls that, otherwise it calls `Sidekiq.default_configration.redis_info`.

It's not massively clear these healthchecks are actually being used for anything since this has been failing on production applications for a couple of weeks without anyone noticing.

We're in the process of speaking to the platform team to see if they're actually being used for anything and if not whether they should be removed.

But for now i'm just updating it to work with Sidekiq 7.

### DIFF
--- a/lib/govuk_app_config/govuk_healthcheck/sidekiq_redis.rb
+++ b/lib/govuk_app_config/govuk_healthcheck/sidekiq_redis.rb
@@ -5,7 +5,14 @@ module GovukHealthcheck
     end
 
     def status
-      Sidekiq.redis_info ? OK : CRITICAL
+      # Sidekiq 7 introduced a default_configuration object which has .redis_info
+      # for querying Redis information. If the default_configuration object isn't present,
+      # we can fall back to the old method of querying it using 'Sidekiq.redis_info'.
+      if Sidekiq.respond_to?(:default_configuration)
+        Sidekiq.default_configuration.redis_info ? OK : CRITICAL
+      else
+        Sidekiq.redis_info ? OK : CRITICAL
+      end
     rescue StandardError
       # One would expect a Redis::BaseConnectionError, but this should be
       # critical if any exception is raised when making a call to redis.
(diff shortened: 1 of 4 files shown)


spec/lib/govuk_healthcheck/sidekiq_redis_spec.rb
```diff
@@ -3,35 +3,68 @@
 require_relative "shared_interface"
 
 RSpec.describe GovukHealthcheck::SidekiqRedis do
-  let(:redis_info) { double(:redis_info) }
-  let(:sidekiq) { double(:sidekiq, redis_info:) }
   before { stub_const("Sidekiq", sidekiq) }
 
-  context "when the database is connected" do
+  context "when the Sidekiq version is >= 7.0.0" do
```
kevindew, line 8: I think this (and the alternate version) should probably explicitly reference `default_configuration` rather than the Sidekiq version - as the test doesn't actually reference that and you have to implicitly understand that setting up default_configuration alludes to it being Sidekiq 7 or not.

kevindew (commented):
Great stuff, just a couple of minor things.

lib/govuk_app_config/govuk_healthcheck/sidekiq_redis.rb
```diff
@@ -5,7 +5,11 @@ def name
     end
 
     def status
-      Sidekiq.redis_info ? OK : CRITICAL
+      if Sidekiq.respond_to?(:redis_info)
```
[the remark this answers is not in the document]
davidgisbey: yea makes sense.

spec/lib/govuk_healthcheck/sidekiq_redis_spec.rb
davidgisbey: cool have updated it and added a comment to the healthcheck above the conditional explaining the issue.

kevindew (approved):
Nice one 👍

PR #444, opened 2025-04-02
Add support for internationalisation of Rails framework strings

hannako:
## Govuk_i18n 

Some languages that are supported by GOV.UK are not yet supported by the
Rails framework. (See [audit](https://docs.google.com/spreadsheets/d/1zbpROLIsFz2ef7JYT8qUlptV7Blp1O0aEENj7CM2HJ8/edit?usp=sharing))

This module allows us to patch these translations so that things like days of the week, month names etc are populated. 

|Before|After|
|-|-|
|Screenshot 2025-04-03 at 13 08 48|Screenshot 2025-04-03 at 13 20 47|

[Trello](https://trello.com/c/pJYPgL2D/3396-prepare-prs-to-import-rails-translations)

See review app at https://github.com/alphagov/frontend/pull/4743 which relies on the temporary commit. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,9 +178,19 @@ app with the following content:
 GovukContentSecurityPolicy.configure
 ```
 
-## Internationalisation rules
+## Internationalisation
 
-Some frontend apps support languages that are not defined in the i18n gem. This provides them with our own custom rules for these languages.
+Some frontend apps support languages that are not yet defined in the [rails-i18n](https://github.com/svenfuchs/rails-i18n) gem that is bundled with Rails.
+
+The missing translations can be added to a locale file in the folder [`lib/govuk_app_config/govuk_i8n`](lib/govuk_app_config/govuk_i18n/). The [`:en`](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en-GB.yml) locale file in the Rails repo  can be used as a guide to key names, as it contains all the keys that Rails uses for a fully internationalised language. It is not necessary to add all of these keys to your file.
+
+See [example.yml](lib/govuk_app_config/govuk_i18n/docs/example.yml) for the most likely structure.
+
+The translations can be enabled in the frontend application by creating a file at `config/initializers/govuk_i18n.rb` with the following content:
+
+```ruby
+GovukI18n.configure
+```
 
 ## Time zone
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.17.0
+
+* Add support for internationalisation of Rails framework strings ([#444](https://github.com/alphagov/govuk_app_config/pull/444))
+
 # 9.16.11
 
 * Update dependencies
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -178,9 +178,20 @@ app with the following content:
...
+The missing translations can be added to a locale file in the folder [`lib/govuk_app_config/govuk_i8n`](lib/govuk_app_config/govuk_i18n/). The [`:en`](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en-GB.yml) locale file in the Rails repo  can be used as a guide to key names, as it contains all the keys that Rails uses for a fully internationalised language. It is not necessary to add all of these keys to your file.
+
+See [example.yml](lib/govuk_app_config/govuk_i18n/docs/example.yml) for the most likely structure.
+
+The translations can be enabled in the frontend application by creating a file at `config/initializers/govuk_i18n.rb` with the following content:
+
+```ruby
+GovukI18n.configure
+```
+``
```
AgaDufrat, line 194:
```suggestion

```

AgaDufrat (approved):
Thanks for a nicely structured PR and including a link to a test app.

I have very little context on this work but did you consider contributing to rails-i18n directly, which we’ve done previously, or raising an issue for this? I appreciate that the timescales may not make those possible options. It looks like Armenian locales were added a few hours ago: https://github.com/svenfuchs/rails-i18n/pull/1144/files within an hour of the PRs being raised which is promising.

Do we need a process in place that will prevent us from having this patch in place, when it’s no longer needed because rails started supporting those translations?

The code itself looks good so I'll give it a tick, not to block this work, but it would be good to have Senior Tech's thoughts on this.

hannako:
Thanks for the review @AgaDufrat 

I think ideally we would contribute to the rails i18n gem. However, there are lot translations required for a fully populated Rails locale file. My experience of sourcing translations for words defined in our apps and gem in recent months, is that it's time consuming and slow. Strings that include interpolation have taken some back and forth. This has lead me to the (anecdotal) opinion that storing a subset of the essential translations i.e. the ones we actually might use in the frontend apps is probably the best compromise.

I like the idea of some logging to alert us to translations that are now available from Rails. I will add that as a follow up PR if that's ok with you.

I'll tag @richardTowers as you've mentioned wanting the opinion of senior tech, and he has some context on this work.

richardTowers:
Yeah, I think that sounds pragmatic.

The situation as I understand it:

- We'll provide partial support for some languages that aren't supported by rails-i18n in this gem
- Partial, because we only need a subset of rails' full i18n support on GOV.UK, and we can't commit to doing all of the extra work to translate the remainder that we don't need
- The code in this repo is licenced compatibly with rails-i18n, and they're welcome to take anything here as a starting point for other locales

I guess we could drop an issue on the upstream project letting them know we've done that, and that they're welcome to our code if they want it?

lib/govuk_app_config/govuk_i18n/zh.yml
```diff
@@ -0,0 +1,113 @@
+---
+zh:
+  date:
+    abbr_day_names:
+    - Sun-zh
```
leenagupte, line 5: Very clever!

leenagupte (approved):
This sounds like a good plan to me. Especially with the alerting being added in a follow-up PR.

Thanks for updating the documentation too 🎉

PR #599, opened 2026-04-16
Prevent double initialisation of Logstasher

sihugh:
Apps that configure their own custom fields can now inadvertently register two listeners, since [the last change which fixed a boot race condition](https://github.com/alphagov/govuk_app_config/commit/f130c88be5b384c4c7891abce9503123efc4fac1).

This change adds a check to see if the configuration has already run. The [app logstasher initialiser](https://github.com/alphagov/content-store/blob/main/config/initializers/logstasher.rb#L2) runs before the [gem's railtie after_initialize hook](https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/railtie.rb#L32), so this should now work correctly.Prevent double initialisation

Apps that configure their own custom fields are currently able to inadvertently register two listeners, since [the last change which fixed a boot race condition](https://github.com/alphagov/govuk_app_config/commit/f130c88be5b384c4c7891abce9503123efc4fac1).

This change adds a check to see if the configuration has already run. The [app logstasher initialiser](https://github.com/alphagov/content-store/blob/main/config/initializers/logstasher.rb#L2) runs before the [gem's railtie after_initialize hook](https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/railtie.rb#L32), so this should now work correctly and only register once, and include custom fields if some are supplied.

Example of dual logging:
Screenshot 2026-04-16 at 17 31 06

## Also add a request_uri logging field to Rails Json logs

This matches the Nginx request_uri field, and includes the query string, which the current path log field does not.

This should make it easier to search for `request_uri: \/search\/all*` and get everything, rather than just Nginx logs

### DIFF
--- a/lib/govuk_app_config/govuk_json_logging.rb
+++ b/lib/govuk_app_config/govuk_json_logging.rb
@@ -56,6 +56,7 @@ module GovukJsonLogging
       # Mirrors Nginx request logging, e.g. GET /path/here HTTP/1.1
       fields[:request] = "#{request.request_method} #{request.fullpath} #{request.headers['SERVER_PROTOCOL']}"
 
+      fields[:request_uri] = request.fullpath
       fields[:govuk_request_id] = request.headers["GOVUK-Request-Id"]
       fields[:varnish_id] = request.headers["X-Varnish"]
       fields[:govuk_app_config] = GovukAppConfig::VERSION
--- a/lib/govuk_app_config/govuk_json_logging.rb
+++ b/lib/govuk_app_config/govuk_json_logging.rb
@@ -16,6 +16,10 @@ module GovukJsonLogging
   end
 
   def self.configure(&block)
+    return if @configured
+
+    @configured = true
+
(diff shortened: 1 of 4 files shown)


lib/govuk_app_config/govuk_json_logging.rb
```diff
@@ -16,6 +16,10 @@ def add_custom_fields(&block)
   end
 
   def self.configure(&block)
+    return if @configured
+
+    @configured = true
```
yndajas, line 21: question: is there a reason for setting this before doing the actual configuration?

yndajas (approved):

lib/govuk_app_config/govuk_json_logging.rb
sihugh: Just my preference for keeping that little bit of logic close together. I accept that putting it at the bottom might be more meaningful!

